PR #1, opened 2020-02-10
Partial Grammar atd using atdgen and yojson

ulziibay:
Lets use Yojson + Atdgen which gives the AST types def in addition to the serialization/deserialization.
Basing the implementation of JSON parsing from https://dev.realworldocaml.org/json.html

Testing:

1. atdgen -t lib/grammar.atd && atdgen -j lib/grammar.atd 
2. Dune build 
3. ./_build/default/bin/main_codegen.exe -parse_grammar tests/arithmetic/grammar.json

TODO:
- [ ] Resolve string decoding issue with Yojson
- [ ] Undo grammar.json changes after adding complex (non-terminal nodes) syntaxes for rules


lib/grammar.atd
```diff
@@ -0,0 +1,42 @@
+type pattern_rule = {
```
ulziibay: @aryx this is the main file

aryx:
Not convinced this is the right approach.
Reading the grammar.json is file should be pretty easy even with json-wheel. Brining atdgen seems overkill to me, but if you have fun with it, go for it.
Let me know when you parse grammar.json and generate from it an AST defined in ast_grammar.re.

ulziibay:
Closing in favor of https://github.com/returntocorp/ocaml-tree-sitter/pull/2

PR #2, opened 2020-02-11
Parse grammar.json manually, and refine ast_grammar to handle Java and Ruby

aryx:
Test plan:

pad@yrax:~/github/ocaml-tree-sitter$ ./_build/default/bin/main_codegen.exe -parse_grammar tests/arithmetic/grammar.json
("program",
 [("program",
   (REPEAT
      (CHOICE
         [(SYMBOL "assignment_statement"); (SYMBOL "expression_statement")])));
   ("assignment_statement",
    (SEQ
       [(SYMBOL "variable"); (STRING "="); (SYMBOL "expression");
         (STRING ";")]));
   ("expression_statement", (SEQ [(SYMBOL "expression"); (STRING ";")]));
   ("expression",
    (CHOICE
       [(SYMBOL "variable"); (SYMBOL "number");
         (SEQ [(SYMBOL "expression"); (STRING "+"); (SYMBOL "expression")]);
         (SEQ [(SYMBOL "expression"); (STRING "-"); (SYMBOL "expression")]);
         (SEQ [(SYMBOL "expression"); (STRING "*"); (SYMBOL "expression")]);
         (SEQ [(SYMBOL "expression"); (STRING "/"); (SYMBOL "expression")]);
         (SEQ [(SYMBOL "expression"); (STRING "^"); (SYMBOL "expression")])]));
   ("variable", (PATTERN "\\a\\w*")); ("number", (PATTERN "\\d+"));
   ("comment", (PATTERN "#.*"))])

Tested also on the other grammar.json under tests/ (for Java, Csharp, Ruby)

### DIFF
--- a/lib/ast_grammar.re
+++ b/lib/ast_grammar.re
@@ -1,26 +1,34 @@
 /* Essence of grammar.json content */
 
+[@deriving show]
 type ident = string;
 
+[@deriving show { with_path : false} ]
 type rule_body =
  /* composite (nodes) */
  | REPEAT(rule_body)
+ | REPEAT1(rule_body)
  | CHOICE(list(rule_body))
  | SEQ(list(rule_body))
+ | ALIAS(rule_body, ident) /* used in Java for AlIAS(identifier, type_identifier) */
  /* atomic (leaves) */
  | SYMBOL(ident)
  | STRING(string)
  | PATTERN(string)
  | TOKEN /* no need to look under that */
+ | IMMEDIATE_TOKEN /* used in Ruby for keyword_parameter, IMMEDIATE_TOKEN(":") */
  | BLANK /* usually used in a CHOICE(...,BLANK) to encode optionality */
 ;
 
-
+[@deriving show]
 type rule = (ident, rule_body);
 
+[@deriving show]
 type rules = list(rule);
 
+[@deriving show]
 type grammar = (ident /* entry point, first rule */, rules);
 
 /* alias */
+[@deriving show]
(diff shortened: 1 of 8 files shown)


aryx:
I'm also learning Reason doing so! Not that simple actually when coming from OCaml.
I already reported a bug to github.com/facebook/reason

aryx:
(note that the -parse_grammar output is better than what github displays above; the result is indented to better show the structure, try it.

lib/ast_grammar.re
```diff
@@ -1,26 +1,34 @@
 /* Essence of grammar.json content */
 
+[@deriving show]
```
ulziibay: what does @deriving show do?

lib/parse_grammar.re
```diff
@@ -12,9 +12,73 @@
...
 /*****************************************************************************/
 /* Entrypoint */
 /*****************************************************************************/
-let parse = _file => raise(Todo);
+let parse = file => {
+    let json = Json_io.load_json(file);
+    switch(json) {
+    | J.Object(xs) => {
+            let rules = List.assoc("rules",xs);
+            parse_grammar(rules)
```
ulziibay: I was missing this syntax. Thanks for this

lib/ast_grammar.re
aryx: It auto generates the Ast_grammar.show_xxx function, a bit like what console.log(anything) can do in Javascript. I use this show_xxx function in test_tree_sitter.re

PR #3, opened 2020-02-11
ci: Add bento check for pull requests

underyx:
Heya 👋

I just made https://github.com/returntocorp/bento-action which will mark PRs as failed if they introduce any new issues. We wanted to dogfood it and test it on all r2c projects, so I'd appreciate a merge 🙃

### DIFF
--- a/.github/workflows/bento.yml
+++ b/.github/workflows/bento.yml
@@ -1,0 +1,13 @@
+name: Bento
+on: [pull_request]
+jobs:
+  bento:
+    runs-on: ubuntu-latest
+    name: Bento checks
+    steps:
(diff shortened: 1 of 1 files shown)


aryx:
Sure, thx!

aryx:
There is no code that bento currently handles in this repository (It's all ocaml/reason code),
or js test files under tests/, so not sure bento can do much.

underyx:
Yep, that's by design. Luke's idea was that every repo should have bento
regardless as the action will auto-update and at some point just magically
start working.

I worry it might be annoying once that happens to have a failure reported,
but let's see how it goes anyway! 🙃

PR #4, opened 2020-02-12
Recursive normalization

ulziibay:
Normalization between A => B, where A is Ast_grammar and B is Ast_grammar_normalized.
Added simple tests for
```
        CHOICE
        /    |
      SYMBOL(A)  CHOICE
            /    |
          SYMBOL(B)  SYMBOL(C)
```
and 
```
              SEQ
             / | \
      CHOICE TOKEN SYMBOL(A)
       /  \
  SYMBOL(B) TOKEN
```

### DIFF
--- a/lib/normalize_grammar.re
+++ b/lib/normalize_grammar.re
@@ -1,6 +1,69 @@
 open Common;
 
-let normalize_body = (rule_body) => (rule_body);
+let normalize_to_simple = body => {
+  switch (body) {
+  | Ast_grammar.TOKEN =>
+    Ast_grammar_normalized.ATOM(
+      Ast_grammar_normalized.TOKEN
+    );
+  | Ast_grammar.SYMBOL(name)=>
+    Ast_grammar_normalized.ATOM(
+      Ast_grammar_normalized.SYMBOL(name)
+    );
+  /* Ignore rest */
+  | _ => {
+    print_string("Inside normalize_to_simple\n");
+    raise(Todo);
+  }
+  }
+}
+
+let normalize_to_atom = body => {
+  switch (body) {
+  | Ast_grammar.TOKEN =>
+    Ast_grammar_normalized.TOKEN
+  | Ast_grammar.SYMBOL(name)=>
+    Ast_grammar_normalized.SYMBOL(name)
+  /* Ignore rest */
+  | _ => {
+    print_string("Inside normalize_to_atom\n");
+    raise(Todo);
+  }
+  }
+}
+
+let normalize_body = (rule_body) => {
+  switch(rule_body) {
+  | Ast_grammar.TOKEN =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.TOKEN
+      )
+    );
+  | Ast_grammar.SYMBOL(name) =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.SYMBOL(name)
+      )
+    );
+  | Ast_grammar.SEQ(bodies) =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.SEQ(
+        List.map(normalize_to_atom, bodies)
+      )
+    );
+  | Ast_grammar.CHOICE(bodies) =>
+    Ast_grammar_normalized.CHOICE(
+      List.map(normalize_to_simple, bodies)
+    );
+  | _ =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.TOKEN
+      )
+    );
+  }
+}
 
 let normalize_rule = ((name, rule_body)) => (name, normalize_body(rule_body));
 
@@ -8,11 +71,8 @@ let normalize_rules = xs => List.map(normalize_rule, xs);
 
 let normalize = ast => {
   switch(ast) {
-  | (_, rules) => {
-      pr2(spf("Length of rules %s \n", string_of_int(List.length(rules))));
-      let normalized_rules = normalize_rules(rules);
-      pr2(spf("Length of normalized rules %s \n", string_of_int(List.length(normalized_rules))));
-      raise(Todo);
+  | (name, rules) => {
+      (name, normalize_rules(rules));
     }
   }
 }
--- a/lib/normalize_grammar.re
+++ b/lib/normalize_grammar.re
@@ -1,10 +1,10 @@
 module A = Ast_grammar;
 module B = Ast_grammar_normalized;
 
+let counter = ref(0);
 let gensym = () => {
-  /* TODO make it static variable */
-  let static_ident_counter = 1;
-  "intermediate" ++ string_of_int(static_ident_counter)
+  incr(counter);
+  "intermediate" ++ string_of_int(counter^);
 }
 
 let rec normalize_to_atom = body => {
--- a/lib/normalize_grammar.re
+++ b/lib/normalize_grammar.re
@@ -1,78 +1,62 @@
-open Common;
+module A = Ast_grammar;
+module B = Ast_grammar_normalized;
 
-let normalize_to_simple = body => {
+let gensym = () => {
+  /* TODO make it static variable */
+  let static_ident_counter = 1;
+  "intermediate" ++ string_of_int(static_ident_counter)
+}
+
+let rec normalize_to_atom = body => {
   switch (body) {
-  | Ast_grammar.TOKEN =>
-    Ast_grammar_normalized.ATOM(
-      Ast_grammar_normalized.TOKEN
-    );
-  | Ast_grammar.SYMBOL(name)=>
-    Ast_grammar_normalized.ATOM(
-      Ast_grammar_normalized.SYMBOL(name)
-    );
-  /* Ignore rest */
+  | A.TOKEN => (B.TOKEN, [])
+  | A.SYMBOL(name)=> (B.SYMBOL(name), [])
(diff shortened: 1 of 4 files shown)


lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
 open Common;
 
-let normalize = _ast =>
-  raise (Todo);
-  
\ No newline at end of file
+let normalize_to_simple = body => {
```
aryx: You can use module aliases to be less verbose:
module A = Ast_grammar;
module B = Ast_grammar_normalized;

so you can then just write A.TOKEN => ... B.TOKEN which is shorter.

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+      )
+    );
+  }
+}
+
+let normalize_rule = ((name, rule_body)) => (name, normalize_body(rule_body));
+
+let normalize_rules = xs => List.map(normalize_rule, xs);
+
+let normalize = ast => {
```
aryx: you can use the pattern above instead of introducing the switch with just one case:

let normalize = ((name, rules)) => ...
Note that extra parenthesis, to disambiguate between 1 parameter that is a pair, with 2 parameters.

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.TOKEN
+      )
+    );
+  }
+}
+
+let normalize_rule = ((name, rule_body)) => (name, normalize_body(rule_body));
+
+let normalize_rules = xs => List.map(normalize_rule, xs);
```
aryx: normalize_rule will actually introduce new rules (for the intermediate nodes), so each normalize_rule
will actually return a list of rules, so you'll need List.flatten(List.map(normalize_rule,xs))

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+  | _ =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.TOKEN
+      )
+    );
+  }
+}
+
+let normalize_rule = ((name, rule_body)) => (name, normalize_body(rule_body));
```
aryx: normalize_body will not just return a normalized_body, it will return also new rules for the intermediates, so you'll need something like

let (this_body, intermediates) = normalize_body(rule_body);
let extra_rules = <do_something_on_intermediates_like_calling_again_normalize_rule_on_it>;
(this_body, extra_rules)

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+  | _ => {
+    print_string("Inside normalize_to_atom\n");
+    raise(Todo);
+  }
+  }
+}
+
+let normalize_body = (rule_body) => {
+  switch(rule_body) {
+  | Ast_grammar.TOKEN =>
```
aryx: With the right module aliasing it can be more readable such as
| A.TOKEN => B.SIMPLE(B.ATOM(B.TOKEN))

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+        Ast_grammar_normalized.TOKEN
+      )
+    );
+  | Ast_grammar.SYMBOL(name) =>
+    Ast_grammar_normalized.SIMPLE(
+      Ast_grammar_normalized.ATOM(
+        Ast_grammar_normalized.SYMBOL(name)
+      )
+    );
+  | Ast_grammar.SEQ(bodies) =>
```
aryx: here you can get intermediates, so you'll need
let xs = List.map(normalize_to_atom, bodies);
let atoms = List.map(fst, xs);
let intermediates = List.flatten(List.map(snd, xs));
(B.SEQ(atoms), intermediates)

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+}
+
+let normalize_to_atom = body => {
+  switch (body) {
+  | Ast_grammar.TOKEN =>
+    Ast_grammar_normalized.TOKEN
+  | Ast_grammar.SYMBOL(name)=>
+    Ast_grammar_normalized.SYMBOL(name)
+  /* Ignore rest */
+  | _ => {
```
aryx: in this case, you'll need to generate a fresh identifier, to introduce a new SYMBOL with this
body to be processed later on by normalize_to_rule_body.

| _ => {
   let fresh_ident = gensym();
   B.SYMBOL(fresh_ident), [fresh_ident, body]

lib/normalize_grammar.re
```diff
@@ -1,5 +1,78 @@
...
+    print_string("Inside normalize_to_simple\n");
+    raise(Todo);
+  }
+  }
+}
+
+let normalize_to_atom = body => {
+  switch (body) {
+  | Ast_grammar.TOKEN =>
+    Ast_grammar_normalized.TOKEN
```
aryx: here there is no need to create intermediate, so you can return the empty list:
| A.TOKEN -> B.TOKEN, []

lib/normalize_grammar.re
ulziibay: Ok.

lib/normalize_grammar.re
ulziibay: ok.

lib/normalize_grammar.re
ulziibay: ok.

lib/normalize_grammar.re
ulziibay: Ok

lib/normalize_grammar.re
ulziibay: ok.

lib/normalize_grammar.re
ulziibay: Ok.

lib/normalize_grammar.re
ulziibay: Ok.

lib/normalize_grammar.re
ulziibay: I addressed it for A.SEQ, and I did something very similar for A.CHOICE matching. LMK if it ok

lib/normalize_grammar.re
```diff
@@ -1,5 +1,62 @@
-open Common;
+module A = Ast_grammar;
+module B = Ast_grammar_normalized;
 
-let normalize = _ast =>
-  raise (Todo);
-  
\ No newline at end of file
+let gensym = () => {
+  /* TODO make it static variable */
```
aryx: let counter = ref(0);
let gensym = () {
   incr(counter);
  "intermediate" ++ string_of_int(counter^);
}

lib/normalize_grammar.re
```diff
@@ -1,5 +1,62 @@
...
+}
+
+let rec normalize_to_atom = body => {
+  switch (body) {
+  | A.TOKEN => (B.TOKEN, [])
+  | A.SYMBOL(name)=> (B.SYMBOL(name), [])
+  /* Create intermediate symbol */
+  | _ => {
+    let fresh_ident = gensym();
+    (B.SYMBOL(fresh_ident), [(fresh_ident, fst(normalize_body(body)))]);
```
aryx: I don't think this will work, you can't ignore snd(normalize_body(body));

In fact I think you should return a Ast_grammar.rule, not an Ast_Grammar_normalized.rule, and in the caller
call recursively the function on it.

lib/normalize_grammar.re
```diff
@@ -1,5 +1,62 @@
...
+    }
+  }
+}
+and normalize_body = (rule_body) => {
+  switch(rule_body) {
+  | A.TOKEN =>
```
aryx: You redo part of the work done in normalize_to_atom.
You can simply call normalize_to_atom for those cases.

lib/normalize_grammar.re
```diff
@@ -1,5 +1,62 @@
...
+      []
+    );
+  | A.SEQ(bodies) => {
+      let xs = List.map(normalize_to_atom, bodies);
+      let atoms = List.map(fst, xs);
+      let intermediates = List.flatten(List.map(snd, xs));
+      (B.SIMPLE(B.SEQ(atoms)), intermediates)
+    };
+  | A.CHOICE(bodies) => {
+      let xs = List.map(normalize_to_atom, bodies);
```
aryx: Here you should call normalize_to_simple, because B.CHOICE take simple, not atoms.
You miss the opportunity to get some CHOICE(SEQ), you'll always get CHOICE(B.ATOM(, which would lead to
less elegant ASTs.

lib/normalize_grammar.re
```diff
@@ -1,5 +1,62 @@
...
+    (
+      B.SIMPLE(B.ATOM(B.TOKEN)),
+      []
+    );
+  }
+}
+
+let normalize_rule = ((name, rule_body)) => {
+  let (this_body, intermediates) = normalize_body(rule_body);
+  let extra_rules = intermediates;
```
aryx: instead you should call normalize_rule on intermediates until there is no more intermediates.

lib/test_tree_sitter.re
```diff
@@ -12,19 +12,121 @@
...
+          B.ATOM(B.SYMBOL("B")),
+          B.ATOM(B.TOKEN),
+        ])
+      )
+    ]
+  );
+  _test_normalization(grammar, expected);
+}
+
+let test_normalization_1 = _ => {
```
aryx: nice.

lib/normalize_grammar.re
ulziibay: ok.

lib/normalize_grammar.re
ulziibay: ok

PR #7, opened 2020-02-14
Handle REPEAT

ulziibay:
As titled. lets handle repeat node by calling `normalize_to_simple`


ulziibay:
@aryx how do we differentiate REPEAT vs REPEAT1 in ast_grammar_normalized?

PR #8, opened 2020-02-14
Handle BLANK as B.OPTION

ulziibay:
Since CHOICE(...,BLANK) is used to encode optionality, let create B.OPTION node in that case

### DIFF
--- a/lib/normalize_grammar.re
+++ b/lib/normalize_grammar.re
@@ -9,7 +9,11 @@ let gensym = () => {
 
 let rec normalize_to_simple = (body: A.rule_body): (B.simple, list(A.rule)) => {
   switch (body) {
+<<<<<<< HEAD
   | A.TOKEN | A.IMMEDIATE_TOKEN | A.BLANK | A.SYMBOL(_) | A.STRING(_) | A.PATTERN(_) => {
+=======
+  | A.TOKEN | A.SYMBOL(_) => {
+>>>>>>> c34f70b... rebase fixes
     let (atom, rest) = normalize_to_atom(body);
     (B.ATOM(atom), rest)
   }
--- a/lib/normalize_grammar.re
+++ b/lib/normalize_grammar.re
@@ -57,8 +57,21 @@ and normalize_body = (rule_body: A.rule_body): (B.rule_body, list(A.rule)) => {
       let xs = List.map(normalize_to_simple, bodies);
       let simples = List.map(fst, xs);
       let intermediates = List.flatten(List.map(snd, xs));
-      (B.CHOICE(simples), intermediates)
-    };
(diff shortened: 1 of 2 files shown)


lib/normalize_grammar.re
```diff
@@ -76,7 +76,21 @@ and normalize_body = (rule_body) => {
...
+        let (_, first) = List.nth(intermediates, 0);
+        switch(first) {
+          | A.BLANK => {
+            (B.OPTION(List.nth(simples, 0)), [])
+          };
+          | _ => (B.CHOICE(simples), intermediates)
+        }
+      } else {
+        (B.CHOICE(simples), intermediates)
+      }
```
ulziibay: @aryx any idea on how to simplify this?

lib/normalize_grammar.re
```diff
@@ -76,7 +76,21 @@ and normalize_body = (rule_body) => {
       let xs = List.map(normalize_to_simple, bodies);
```
aryx: you should check for the presence of BLANK here, on bodies, before the normalize_to_simple calls.
BLANK is always the last element of CHOICE, so you can do
| A.CHOICE(bodies) => {
  switch (List.rev(bodies)) {
 | [BLANK] -> failwith("Impossible, a single BLANK in a CHOICE")
 | [BLANK, other_body] ->
    let (simple, rest) = normalize_to_simple(other_body);
    (B.OPTION(simple), rest)
| [BLANK,  ...other_bodies] ->
  /* this will create a gensymed ATOM(SYMBOL) for this CHOICE */
  let (simple, rest) = normalize_to_simple(A.CHOICE(other_bodies));
  (B.OPTION(simple), rest)

ulziibay: Ok addressed

lib/normalize_grammar.re
```diff
@@ -33,9 +29,7 @@ let rec normalize_to_simple = (body: A.rule_body): (B.simple, list(A.rule)) => {
 and normalize_to_atom = (body: A.rule_body): (B.atom, list(A.rule)) => {
   switch (body) {
   | A.TOKEN => (B.TOKEN, [])
-  | A.SYMBOL(name) => (B.SYMBOL(name), [])
-  | A.STRING(name) => (B.SYMBOL(name), [])
-  | A.PATTERN(name) => (B.SYMBOL(name), [])
+  | A.SYMBOL(name) | A.STRING(name) | A.PATTERN(name) => (B.SYMBOL(name), [])
```
aryx: no, a symbol and a string are very different.
STRING AND PATTERN should convert to TOKEN I think.

ulziibay: ok

PR #24, opened 2020-02-20
Codegen type constructor and overall program type

ulziibay:
Noticed the absence of the type constructors from the 
AST files. This adds them back so that we can use them
for the json codegen code.


tests/java/ast_java.re
```diff
@@ -564,3 +564,314 @@ and inferred_parameters = (string, intermediate7, string)
 and enum_body_declarations = (string, intermediate93)
 and synchronized_statement = (string, parenthesized_expression, block)
 and program = list(_statement);
+type cst_node =
```
ulziibay: @aryx i feel that there is too many duplicates caused by the fact that I create new type for each "CHOICE" i find in the grammar. This is hitting 
```
Too many non-constant constructors
     -- maximum is 246 non-constant constructors
```
I believe I have to have hierarchical type constructor definition here. Then the simple concat all the rule strings with "and" does not work for generating the types and their constructor!

lib/codegen_types.re
```diff
@@ -65,39 +73,55 @@ let codegen_rule_body = (rule_body: B.rule_body): string => {
...
-            | B.SIMPLE(B.ATOM(B.TOKEN)) => -1
-            | _ => {
-               switch(b_body) {
-               | B.SIMPLE(B.ATOM(B.TOKEN)) => 1
-               | _ => String.compare(a_type, b_type)
-               }
-            }
-            }
+let sort_by_type = ((a_i, a_body): B.rule, (b_i, b_body): B.rule) => {
+   if (a_i == "program") {
```
aryx: You should use instead the name of the entry point (it's the first element in ast_grammar_normalized.grammar)

lib/codegen_types.re
```diff
@@ -65,39 +73,55 @@ let codegen_rule_body = (rule_body: B.rule_body): string => {
...
+   let rules_sorted_by_type = List.sort(sort_by_type, rules);
    List.map(((name, body): B.rule) => {
-      wrap_ident(name) ++ " = " ++ codegen_rule_body(body)
+      (name, wrap_ident(name) ++ " = " ++ codegen_rule_body(body))
    }, rules_sorted_by_type)
 }
 
-let codegen = ((_, rules): B.grammar): (string) => {
+let codegen_rule_constructor = (rule_names: list(string)): string => {
+   "type cst_node = \n" ++
```
aryx: Why do you need that? What is this cst_node for?

tests/arithmetic/ast_arithmetic.re
```diff
@@ -16,3 +16,13 @@ and expression =
 and expression_statement = (expression, string)
 and assignment_statement = (variable, string, expression, string)
 and program = list(intermediate1);
+type cst_node =
```
aryx: there is no need for that.

tests/java/ast_java.re
aryx: There is no need for this cst_node, so this problem should go away.

ulziibay:
Gonna rethink this!

PR #27, opened 2020-02-24
Helper script that installs treesitter, generates bindings and produces cst  for a given file in a specified language

ulziibay:
Run with
1) ./lib/generate_cst.py tests/arithmetic/grammar.js tests/arithmetic/test.arith > test.arith.cst.json

### DIFF
--- a/lib/generate_cst.py
+++ b/lib/generate_cst.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+import argparse
+import json
+import os
+import subprocess
+import sys
+from typing import Any, Dict, List, Optional
+from json import JSONDecodeError
+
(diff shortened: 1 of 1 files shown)


lib/generate_cst.py
```diff
@@ -0,0 +1,162 @@
...
+var args = process.argv.slice(2);
+
+var fs = require("fs");
+const sourceCode = fs.readFileSync(args[0]).toString();
+
+const parser = new Parser();
+parser.setLanguage({language_name.capitalize()});
+const tree = parser.parse(sourceCode);
+
+console.log(JSON.stringify(tree.rootNode, ["type", "children"], 2))
```
aryx: can put a TODO here or a comment that later we will add startPOS, etc. but for now
we simplify things.

PR #29, opened 2020-02-25
Make Token of NAST of some string

ulziibay:
When we have STRING nodes in the AST, lets normalize them 
to TOKEN(some(string)) so that we keep the distinction between
certain types


ulziibay:
# normalized grammar
```
("program",
 [("assignment_statement",
   (SIMPLE
      (SEQ
         [(SYMBOL "variable"); (TOKEN (Some "=")); (SYMBOL "expression");
           (TOKEN (Some ";"))])));
   ("comment", (SIMPLE (ATOM (TOKEN None))));
   ("expression",
    (CHOICE
       [(ATOM (SYMBOL "variable")); (ATOM (SYMBOL "number"));
         (SEQ
            [(SYMBOL "expression"); (TOKEN (Some "+")); (SYMBOL "expression")
              ]);
         (SEQ
            [(SYMBOL "expression"); (TOKEN (Some "-")); (SYMBOL "expression")
              ]);
         (SEQ
            [(SYMBOL "expression"); (TOKEN (Some "*")); (SYMBOL "expression")
              ]);
         (SEQ
            [(SYMBOL "expression"); (TOKEN (Some "/")); (SYMBOL "expression")
              ]);
         (SEQ
            [(SYMBOL "expression"); (TOKEN (Some "^")); (SYMBOL "expression")
              ])
         ]));
   ("expression_statement",
    (SIMPLE (SEQ [(SYMBOL "expression"); (TOKEN (Some ";"))])));
   ("intermediate1",
    (CHOICE
       [(ATOM (SYMBOL "assignment_statement"));
         (ATOM (SYMBOL "expression_statement"))]));
   ("number", (SIMPLE (ATOM (TOKEN None))));
   ("program", (REPEAT (ATOM (SYMBOL "intermediate1"))));
   ("variable", (SIMPLE (ATOM (TOKEN None))))])%    
```

lib/ast_grammar_normalized.re
```diff
@@ -9,7 +9,7 @@ type ident = string;
 [@deriving show { with_path : false} ]
 type atom =
  | SYMBOL(ident) /* codegen: ident */
- | TOKEN
+ | TOKEN(option(ident))
```
aryx: no, option(string).
They are the same types effectively (one is a synonymous of the other), but they do not denote the same conceptual
thing.
";" is not the same than "assignement_expression"

aryx: maybe we could keep TOKEN as is but add STRING(string) like in ast_grammar.re

lib/codegen_types.re
```diff
@@ -15,7 +15,8 @@ let wrap_ident= (ident:string) : string => {
 }
 let codegen_atom = (atom: B.atom): string => {
    switch(atom) {
-   | B.TOKEN => "string" /* tokens are string */
+   | B.TOKEN(None) => "string" /* tokens are empty */
+   | B.TOKEN(Some(name)) => wrap_ident(name)
```
aryx: so, this is wrong. it should be something like "token" that we could define to be
a synonymous for "string" in the generated type file.

lib/normalize_grammar.re
```diff
@@ -28,7 +28,8 @@ let rec normalize_to_simple = (body: A.rule_body): (B.simple, list(A.rule)) => {
 }
 and normalize_to_atom = (body: A.rule_body): (B.atom, list(A.rule)) => {
   switch (body) {
-  | A.TOKEN |  A.IMMEDIATE_TOKEN | A.BLANK | A.STRING(_) | A.PATTERN(_) => (B.TOKEN, [])
+  | A.TOKEN |  A.IMMEDIATE_TOKEN | A.BLANK |A.PATTERN(_) => (B.TOKEN(None), [])
+  | A.STRING(name) =>  (B.TOKEN(Some(name)), [])
```
aryx: A.STRING(str) => (B.TOKEN(Some(str)), [])

PR #38, opened 2020-05-20
Multiple changes, generated code for Ruby compiles

mjambon:
This PR includes multiple fixes and improvements. I'll try to make these PRs shorter in the future, with a preference for sending them out at the end of my day for our night shift personnel.

This branch provides:

* recovery of `optional` rules, which tree-sitter turns into choice(thing, blank) when converting to grammar.json. Uses an option type in ocaml.
* several fixes in code generation having to do with managing the sequence of matchers.
* improved readability/debuggability of generated code:
  * added a Paren construct to the pretty-printing tree (`Indent` module) for more compact output. `(parse_something)` is now one line instead of three.
  * plugged-in the topological sorting of groups of mutually-dependent definitions.

The motivation for the last two sub-bullets was to make it easier to debug the generated code. It did help a bit. Type error messages are still scary (I would say it's due to input being parsed into nested pairs, which are polymorphic i.e. `type ('a, 'b) pair = 'a * 'b`). Hopefully we'll hit such errors less and less.

I'll leave a few inline comments in the PR.

Internal doc: https://returntocorp.quip.com/mDY9AmeBhEQL/Tree-sitter-integration-into-pfff


src/gen/lib/AST_grammar.ml
```diff
@@ -1,5 +1,5 @@
 (*
```
mjambon, line 1: This file was split into type definitions (still `AST_grammar`) and converter code (`AST_grammar_conv`).

src/gen/lib/Codegen_parse.ml
```diff
@@ -10,6 +10,13 @@ open Indent.Types
 
 let debug = false
 
+(* All rule names and other names directly defined in grammar.json
+   must go through this translation. For example, it turns
+   "true" into "true_" because "true" is a reserved keyword in the generated
+   code.
+*)
+let trans = Codegen_util.translate_ident
```
mjambon, line 18: Fixes name clashes we ran into when testing on the Ruby grammar e.g. `method` is a rule name that can't be used as an OCaml type. This was done (for Reason) for generating AST.ml but not for Parse.ml. It's clean now.

src/gen/lib/Indent.ml
```diff
@@ -1,34 +1,13 @@
...
-   let fmt_binding x =
-     [
-       Line "let x =";
-       Block (fmt_body ...);
-     ]
-
-   ...
-
-   let formatted_code = Indent.to_string (fmt_bindings ...)
 *)
```
mjambon, line 3: moved to mli

tests/ruby/examples/ex1.rb.src
```diff
@@ -1 +1 @@
-print "Hello, World!\n"
\ No newline at end of file
+17
```
mjambon, line 1: Our parser fails on this example atm. That's the next thing I'm working on.

tests/ruby/tree-sitter-ruby.js
```diff
@@ -0,0 +1,38 @@
...
+  Warning: This file is meant to be copied into the current folder.
+  A direct call or a symbolic link won't work due to "require('.')".
+
+  Requires:
+  - tree-sitter, tree-sitter-cli.
+  - the products of running 'tree-sitter generate' in the current folder.
+*/
+
+const Parser = require('tree-sitter');
+const pl = require('tree-sitter-ruby');
```
mjambon, line 15: I copy-pasted the whole file and this is the only line that changed. Will work on packaging parsers cleanly later.

src/gen/lib/AST_grammar.ml
```diff
@@ -15,87 +15,22 @@ type rule_body =
   | Repeat of rule_body
   | Repeat1 of rule_body
   | Choice of rule_body list
+  | Optional of rule_body
```
aryx, line 18: yep, good to infer it back. Actually in the original code we had this ast_grammar_normalized.ml (or .re) where the goal
was to normalize grammar.json in something better. This type had this OPTION too inferred back from CHOICE(....BLANK).

aryx: I also lift CHOICE up, to generate regular variants.
type ident = string
  [@@deriving show]

type atom =
  | SYMBOL of ident 
  | TOKEN 
  | STRING of string 
  [@@deriving show { with_path = false }]

type simple =
  | ATOM of atom 
  | SEQ of simple list 
  | REPEAT of simple | REPEAT1 of simple
  | OPTION of simple 
  [@@deriving show { with_path = false }]

type rule_body =
  | CHOICE of simple list 
  | SIMPLE of simple 
  [@@deriving show { with_path = false }]

type rule = (ident * rule_body)

aryx: Also here: https://github.com/returntocorp/ocaml-tree-sitter/blob/to_menhir/lib/ast_grammar_normalized2.ml

src/gen/lib/AST_grammar_conv.ml
```diff
@@ -0,0 +1,94 @@
...
+*)
+let rec translate (x : Tree_sitter_t.rule_body) =
+  match x with
+  | SYMBOL ident -> Symbol ident
+  | STRING s -> String s
+  | PATTERN s -> Pattern s
+  | BLANK -> Blank
+  | REPEAT x -> Repeat (translate x)
+  | REPEAT1 x -> Repeat1 (translate x)
+  | CHOICE [x; BLANK] -> Optional (translate x)
```
aryx, line 18: I think you can have multiple elements here, not just one 'x'.
See what I wrote there: https://github.com/returntocorp/ocaml-tree-sitter/blob/to_menhir/lib/normalize_grammar2.ml 
and especially this function: https://github.com/returntocorp/ocaml-tree-sitter/blob/c71acb393973081f4f512e2e0a0332a427095e5a/lib/normalize_grammar2.ml#L55

tests/ruby/tree-sitter-ruby.js
aryx: yep, copy-paste is fine for fast prototyping :)

src/gen/lib/AST_grammar.ml
mjambon: I see. Not sure which is better in practice. I like a flat view until it gets unmanageable, like for example if we had 10 types of atoms rather than just 3.

src/gen/lib/AST_grammar_conv.ml
mjambon: In my current implementation, `Blank` can remain, and it maps to `unit`. In the context of a choice, it becomes something like `` `Case5 ()` (although just `` `Case5`` would be nicer).

PR #39, opened 2020-05-21
Aliases

mjambon:
This branch takes care of aliases, of which there are two kinds. This was lightly tested, seems to be working.

I also set up some unit testing, see `src/gen/test`.


aryx:
You can maybe copy text from https://tree-sitter.github.io/tree-sitter/creating-parsers that explains the alias feature.

src/gen/lib/AST_grammar.ml
```diff
@@ -4,10 +4,26 @@
...
+    (* original name of the alias, which may not be unique.
+       This is what's exposed in the tree-sitter parser output. *)
+}
+
 type rule_body =
   (* atomic (leaves) *)
-  | Symbol of ident
+  | Symbol of (ident * alias option)
+     (* reference to a named rule. The optional part is the name shown to us
+        in the parsing result due to an ALIAS. *)
```
aryx, line 21: maybe an example from a grammar.js, like   alias($._literal_contents, $.bare_string) from ruby grammar.js?

src/gen/lib/AST_grammar.ml
```diff
@@ -4,10 +4,26 @@
...
+
   | String of string
+     (* constant string as exposed in the parsing output from tree-sitter.
+        It is either the result of a STRING or an anonymous ALIAS. *)
+
```
aryx, line 26: maybe an example for the anonymous ALIAS, like saying "(as in alias($._binary_minus, '-') in Ruby's grammar.js)"

src/gen/lib/AST_grammar.ml
```diff
@@ -4,10 +4,26 @@
 
 type ident = string
 
+type alias = {
```
aryx, line 7: A comment here would be useful, and an example.
For example with alias($._binary_minus, '-') what will be the value of id and name? (or maybe for those cases
there's no alias?)
And with alias($._literal_contents, $.bare_string)? what are id and name? (or maybe say that this is converted
into Symbol("_literal_contents", Some { id = ??; name = "bare_string")?

mjambon:
I'll document this. Sorry, I tried to rush it last night.

mjambon:
I added documentation about aliases and leaf nodes in `Alias.mli` and `AST_grammar.ml`. Should be clearer now.

aryx:
Excellent documentation. Now it makes sense.

PR #41, opened 2020-05-26
Support hidden rules and extras + tracing + fixes

mjambon:
This version is able to parse some Ruby programs, depending on which syntactic constructs they use.

This branch adds the following:

* support for hidden rules. Hidden rules start with an underscore as in `_method_call`. Instead of a node with that `type` field, the children of the node are inlined. If a node is childless, it disappears completely.
* optional tracing of the parser calls.
* a few bug fixes found with the help of tracing.
* support for the `extras` field of grammar.json. These are symbols like comments that can occur anywhere in the program independently from any grammar rules. ocaml-tree-sitter now removes these nodes in a first pass.

Implements https://github.com/returntocorp/ocaml-tree-sitter/issues/36

We're now hitting problems due to the inlining of so-called hidden rules. This should be solved mostly or completely by editing the Ruby (or other) grammar definition so as to not hide anything, or by adding such option to tree-sitter.


src/gen/lib/AST_grammar_conv.ml
```diff
@@ -44,8 +44,8 @@ let translate find_alias (x : Tree_sitter_t.rule_body) =
     | PREC_RIGHT (_opt_prio, x) -> translate x
     | FIELD (_name, x) -> translate x (* TODO not sure about ignoring this *)
     | ALIAS alias -> translate_alias find_alias alias
-    | IMMEDIATE_TOKEN x -> translate x (* TODO check what this is *)
-    | TOKEN x -> translate x
+    | IMMEDIATE_TOKEN _body -> String "immediate token"
+    | TOKEN _body -> String "token"
```
mjambon, line 48: I think this is incorrect. I'll investigate.

mjambon: This made me discover a few other things I misunderstood about aliases and some bugs we may run into. I tried to simplify things but it's not getting that simpler and it would require a rewrite.

I'm now leaning toward eliminating all support for aliases. It would simplify our program a lot and eliminate bugs. We just have to eliminate aliases from the grammar.js files, which shouldn't be a problem. I was hoping to preserve them to give us more insightful type names in the AST but it's looking like a lost cause.

aryx: I'm all in favor of simplifying the code and our job by adjusting the grammar.js, especially because we will at some point anyway modify those grammars (e.g., to add ellipsis).

src/gen/lib/AST_grammar_conv.ml
```diff
@@ -123,6 +130,13 @@ let tsort_rules find_rule_aliases rules =
       { name; aliases; is_rec; body }) group
   ) sorted
 
+let filter_extras bodies =
+  List.filter_map (fun (x : Tree_sitter_t.rule_body) ->
+    match x with
+    | SYMBOL name -> Some name
+    | _ -> None
```
aryx, line 137: You can have something other than a SYMBOL in extras?

src/run/lib/Combine.ml
```diff
@@ -72,7 +134,12 @@ let rec find_longest_match parse_elt stack nodes =
   | None -> stack
   | Some (elt, remaining_nodes) ->
       let stack = push elt remaining_nodes stack in
-      find_longest_match parse_elt stack remaining_nodes
+      if remaining_nodes == nodes (* physical equality *) then
```
aryx, line 137: nice comment highlighting the use of == (easy to miss indeed).

src/run/lib/Main.ml
```diff
@@ -10,6 +10,7 @@ let safe_run f =
   try f ()
   with e ->
     let trace = Printexc.get_backtrace () in
+    flush stdout;
```
aryx, line 13: I never really understood when we need to flush ... I usually assume if your
printed string has an \n then this will flush it. I am probably wrong.

aryx:
+1 for the tracing and all those nice test examples.

src/gen/lib/AST_grammar_conv.ml
mjambon: Pending confirmation that will come in today, I think we'll be able to remove support for aliases and hidden rules completely.

src/gen/lib/AST_grammar_conv.ml
mjambon: The cases I'm aware of are:

* regexps (comments, whitespace): those do not show up in the json output.
* symbols: those show up in the parsing result, and need to be removed.

I need to check what happens for string constants.

There could also be a construct like a seq(). Since they don't have a name, I presume they don't show up in the output.

src/run/lib/Main.ml
mjambon: Here I make sure that anything to be printed on stdout gets printed before we print something on stderr (or exit with nonzero status, causing the output to be lost) to preserve chronological order.

OCaml's predefined out_channels stdout and stderr don't flush automatically when calling a printf function, in contrast to libc streams (where printf to stdout flushes with each newline, and printf to stderr flushes for each call).

Some ocaml functions do flush stdout: print_newline, print_endline, some read_* functions. Those are tricky to remember.

Explicit flushing with printf functions can be done with `%!` e.g. `Printf.printf "Hello!\n%!"`.

src/gen/lib/AST_grammar_conv.ml
mjambon: I thought STRING constants could be supported in the list of "extras" but they cause a runtime error to the tree-sitter parser. The doc says they must be tokens, and we only care if they have a name, because otherwise they don't show up in the parser's output and there's nothing for us to do.

mjambon:
Thanks for the review! I'm merging this.

PR #43, opened 2020-05-29
Simplifications and fixes

mjambon:
I introduced `lang` folder for holding production languages, as opposed to pure tests in `tests`. The setup used for `lang/ruby` should be completely reusable for other languages in the future, including `lang/ruby/Makefile`.

I made simplifications from two ends:

* Added a script for patching `grammar.js` so as to remove the unsupported or undesirable features from the original `grammar.js` (aliases, hidden rules, supertypes).
* Removed support for aliases in the ocaml code.

I also fixed the handling of the token() construct. The AST now has a node that subsumes the 3 types of tokens (string, pattern, token).

I had success parsing non-trivial Ruby programs. I left one in the repo.

~~Note that the AST dumped from ocaml is unreadable at the moment. See screenshot below.~~
image

I just added some better pretty-printing of the AST using ppx_sexp_conv + my own pretty-printer. Result below.
image


scripts/simplify-grammar
```diff
@@ -0,0 +1,90 @@
+#! /usr/bin/env bash
+#
+# Remove aliases and unhide hidden rules, for easier support by
+# ocaml-tree-sitter.
+#
+# Warning: this is based on fragile regexps. It can easily fail.
+# A good solution would be a system of find-and-replace that's aware of
+# javascript syntax...
```
aryx, line 8: We have this :) it's called spatch, in semgrep/bin/main_spatch.ml

aryx: But sed sounds like a good v1

src/gen/lib/Alias.mli
```diff
@@ -1,225 +0,0 @@
-(* Extract ALIAS nodes and assign them globally-unique names.
-
-   What are aliases in a tree-sitter grammar?
```
aryx, line 3: maybe we could save this comment somewhere, in some doc, explaining what they are, why they cause pain for us,
and how we remove them. Same for the _.

src/gen/lib/Codegen_AST.ml
```diff
@@ -20,6 +20,7 @@ let preamble grammar =
    entrypoint: %s
 *)
 
+open! Sexplib.Conv
```
aryx, line 23: you need the !?

mjambon: I guess I'm doing something wrong. The `open` shouldn't be needed at all.

There's also the problem that the sexp representation loses information about the different kinds of types (variant, tuple, list, option) and I don't want to invest too much time in this. If a better dumper turns out to be needed, we can generate our own code for pretty-printing an AST... This would convert a grammar-specific AST to a generic tree of this type:
```
type t =
| Atom of string
| List of t list
| Variant of string * t list option
| Option of t option
| Tuple of t list
```
and then we'd pretty-print this however we like.

src/gen/lib/Alias.mli
mjambon: Yes. I was waiting a bit for things to settle before writing a general readme for the project. I'll fish these comments out of the git history.

aryx:
Yes, ihave already this. See pfff/commons/ocaml.Ml

PR #44, opened 2020-05-31
Script to scan public projects + better input dumper

mjambon:
This branch adds 2 things:

* From `lang/ruby`, `make stat` starts the work of fetching git projects and running our parser on it, producing a csv file. Current parsing success rate is around 50%. The code should work on other languages than Ruby (just change the `projects.txt` and `extensions.txt` files).
* Input from tree-sitter is dumped in a more readable format than the original json. Looks like this for this `arithmetic` example `ex1`:

```
program              
| expression_statement
|   expression
|   | number
|   ;
| expression_statement
|   expression
|   | expression
|   |   number
|   | +
|   | expression
|   |   number
|   ;
| expression_statement
|   expression
|   | expression
|   |   number
|   | -
|   | expression
|   |   number
|   ;
| expression_statement
|   expression
|   | expression
|   |   number
|   | *
|   | expression
|   |   number
|   ;
| expression_statement
|   expression
|   | expression
|   |   number
|   | /
|   | expression
|   |   number
|   ;
| expression_statement
|   expression
|   | expression
|   |   number
|   | ^
|   | expression
|   |   number
|   ;
```
(vertical bars have no semantic value, they're only for readability; it's more obvious when working on typical, deeply nested programs)

### DIFF
--- a/scripts/lang-stat
+++ b/scripts/lang-stat
@@ -1,0 +1,101 @@
+#! /usr/bin/env bash
+#
+# Fetch git projects from a list and run our parser on the files for
+# a language of interest.
+#
+set -eu -o pipefail
+
+progname=$(basename "$0")
+
+usage() {
+  cat <<EOF
+Usage: $progname PROJECTS_FILE EXTENSIONS_FILE
+
+For each project specified by a git URL in PROJECTS_FILE, run the local
+parser on all the files of the project whose extension matches one of the
+extensions in EXTENSIONS_FILE.
+
+Example: $progname projects.txt extensions.txt
+EOF
+}
+
+error() {
+  echo "Error: $*" >&2
+  exit 1
+}
+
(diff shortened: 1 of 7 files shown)


scripts/lang-stat
```diff
@@ -0,0 +1,101 @@
...
+  ) > tmp/parse.log 2>&1
+}
+
+handle_project() {
+  echo "Fetch $url."
+  (
+    cd tmp
+    rm -rf repo
+    rm -f inputs
+    git clone --depth 1 "$url" repo
```
aryx, line 67: this will be redone each time you do 'make stat'?
might be good to reuse between 2 runs the repo if it's there already no? all those git clone
might take quite some time on my internet connexion.

mjambon: indeed, it's quite slow :-)
Parsing is quite slow too, due to writing big files on persistent storage. I already improved some of it in an upcoming branch.

PR #45, opened 2020-06-01
Fix backtracking

mjambon:
This adds a few tests that were failing on Ruby code, and they now pass.

The major bug was in the generation of the parse_inline_X functions, which used to be correct at some point and I had reverted in a haste. I also discovered a plain bug in the `Combine.parse_optional` function. All this is now tested, reducing the risk of regression.

Parsing now succeeds on 99.5% of `.rb` files found in popular Ruby projects.

Next, I'm working on documentation, simplification, and clean up. Lots of files will change or move, so it's probably better to keep it separate from this PR.


scripts/lang-stat
```diff
@@ -53,7 +53,7 @@ parse() {
     cd tmp
     json=out.json
     ast=out.ast
-    "$tree_sitter_parser" "$input" | ydump > "$json"
+    "$tree_sitter_parser" "$input" > "$json"
     "$ocaml_parser" "$input" "$json" > /dev/null
   ) > tmp/parse.log 2>&1
 }
```
mjambon, line 59: Performance of the stat collection is still IO-bound. A further cheap optimization here would be to write to ramdisk. Hopefully we'll be able to remove the conversion to json completely.

src/gen/lib/Codegen_parse.ml
```diff
@@ -2,6 +2,8 @@
    Code generator for the AST.ml file.
 
    This produces code similar to what's found in ../../run/lib/Sample.ml
+
+   TODO: explain the problem and the solution in a document.
```
mjambon, line 6: This will be in the next PR.

PR #46, opened 2020-06-09
Generate good names for variants

mjambon:
This branch:

* eliminates support for hidden rules (`_foo`), which simplifies the code and tests
* generates good names for variant constructors

The naming is up for review. Here's the generated `AST.ml` for Ruby: https://gist.github.com/mjambon/d87d124cc5c735bc57bd637d766e55c7


src/gen/lib/Abbrev.ml
```diff
@@ -0,0 +1,169 @@
+(*
+   Abbreviations for common words, to be used in generated names.
+   These are meant to be better than truncating a word at an arbitrary
+   offset.
+*)
+
+let list = [
```
aryx, line 7: I like it. Hopefully most grammars will use similar names.

src/gen/lib/Punct.mli
```diff
@@ -0,0 +1,9 @@
+(*
```
aryx, line 1: Great!

src/gen/lib/Abbrev.ml
mjambon: We can always add abbreviations for _new_ words later.

mjambon:
For pretty-printing type definitions, I may end up with the same solution I used for atdgen, using easy-format. The output looks like this:

```ocaml
(* Auto-generated from "Tree_sitter.atd" *)
              [@@@ocaml.warning "-27-32-35-39"]

type ident = string

type alias = { value: string; named: bool; content: rule_body }

and rule_body = 
    SYMBOL of ident
  | STRING of string
  | PATTERN of string
  | BLANK
  | REPEAT of rule_body
  | REPEAT1 of rule_body
  | CHOICE of rule_body list
  | SEQ of rule_body list
  | PREC of (int * rule_body)
  | PREC_DYNAMIC of (int * rule_body)
  | PREC_LEFT of (int option * rule_body)
  | PREC_RIGHT of (int option * rule_body)
  | ALIAS of alias
  | FIELD of (ident * rule_body)
  | IMMEDIATE_TOKEN of rule_body
  | TOKEN of rule_body


type grammar = {
  name: ident;
  rules: (ident * rule_body) list;
  extras: rule_body list;
  inline: ident list;
  conflicts: ident list list;
  externals: rule_body list;
  word: ident option;
  supertypes: ident list
}
```

Other solutions include:
* old camlp4 (now camlp5) used to do a great job a formatting or reformatting ocaml code but I don't really want to depend on it.
* refmt doesn't do a good job with ocaml syntax.
* ocamlformat is conservative and looks like it won't eliminate line feeds.
* I've never found the formatting language of the Format module (and others like Fmt) to be readable or safe (which is why I came up with easy-format).

mjambon:
Thanks for the review. I'm merging this, then I'll improve the compactness of type definitions.

PR #47, opened 2020-06-09
Use easy-format to pretty-print type definitions more compactly.

mjambon:
`AST.ml` for Ruby just went from 2059 to 930 lines.

Before: https://gist.github.com/mjambon/d87d124cc5c735bc57bd637d766e55c7
After: https://gist.github.com/mjambon/d7e4cba40d0f1a6b0324a902204061ca


aryx:
For example we could remove the intermediate
type self = Token.t (* "self" *)
in
type variable = [
    `Self of self
  | `Super of super
  | `Inst_var of instance_variable
  | `Class_var of class_variable
and just have
type variable = [
 `Self of Token.t (* "self" *)
 ...

especially because the variant name is already great.
Same with
and begin_block = (
    Token.t (* "BEGIN" *)
  * Token.t (* "{" *)
  * statements option
  * Token.t (* "}" *)
)
which is used only once in 'statement' in which case I would rather have
and statement = [
  ...
  | `Stmt_begin_blk of Token.t (* "BEGIN" *) * Token.T (* "{" *) * statements option * Token.t (* "}" *)
if then can even fit on a single line of 80 chars

mjambon:
Such inlining would make sense. Let me think about it.

PR #48, opened 2020-06-09
Insert token information

mjambon:
This saves the reader the effort of looking up aliases of `Token.t`.

Before:
```ocaml
type variable = [
    `Self of self
  | `Super of super
  | `Inst_var of instance_variable
  | `Class_var of class_variable
  | `Glob_var of global_variable
  | `Id of identifier
  | `Cst of constant
]
```
After:
```ocaml
type variable = [
    `Self of Token.t (* "self" *)
  | `Super of Token.t (* "super" *)
  | `Inst_var of instance_variable (*tok*)
  | `Class_var of class_variable (*tok*)
  | `Glob_var of global_variable (*tok*)
  | `Id of identifier (*tok*)
  | `Cst of constant (*tok*)
]
```

Before: https://gist.github.com/mjambon/d7e4cba40d0f1a6b0324a902204061ca
After: https://gist.github.com/mjambon/32e1944deca84e76f2baed56cdda6e5a

### DIFF
--- a/src/gen/lib/Codegen_AST.ml
+++ b/src/gen/lib/Codegen_AST.ml
@@ -144,43 +144,64 @@ open Ocaml_tree_sitter_run
     grammar.name
     grammar.entrypoint
 
-let rec format_body body : E.t =
-  match body with
-  | Symbol ident ->
-      Fmt.atom (trans ident)
-  | Token { name = _; description = Constant cst } ->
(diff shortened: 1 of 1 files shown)


aryx:
Also because you inline self, you don't need to keep the self type definition because it's not used anymore.

Issue #51, opened 2020-06-12
Create standalone C/OCaml parsers from tree-sitter grammars


mjambon:
Here, "standalone" means independent from javascript/nodejs. Ideally, we'd produce a library that doesn't call executables and ideally doesn't generate an intermediate json file during parsing.

See internal roadmap at https://returntocorp.quip.com/talEAyuKXeOI/WIP-Tree-Sitter-Milestones

PR #54, opened 2020-06-19
Use tree-sitter's C API

mjambon:
We can now parse programming languages directly from OCaml without going through json.

* Parsing from json is still supported. I thought it could be useful for debugging.
* The C bindings were taken from the reason-tree-sitter project.
* We have a shell script (`gen-ocaml`) that generates bindings for the code generated by tree-sitter and also generates the appropriate dune files.
* Simplification: The `tests` and `lang` folders now use the same makefile and scripts.
* The supported languages (currently just `ruby`) can be built and installed from the `lang` folder with `make && make install`.
* I renamed AST to CST pretty much everywhere since it's more correct for what we generate.

As a user of a generated parser, the idea is that you'd use `Tree_sitter_ruby.Parse.string` for testing things in the toplevel. Otherwise, `Tree_sitter_ruby.Parse.file` is the other function which should be convenient and sufficient for parsing source files.

Implements #51


lang/Makefile.common
```diff
@@ -0,0 +1,68 @@
+#
+# Generic makefile to build and test a parser for a specific language.
+#
+
+PROJECT_ROOT = $(shell git rev-parse --show-toplevel)
```
aryx, line 5: nice! I was always manually doing ../../../

Makefile
```diff
@@ -11,7 +20,7 @@ setup:
 # Keep things like node_modules that are worth keeping around
 .PHONY: clean
 clean:
-	rm -r bin
+	rm -rf bin
```
aryx, line 23: ah yes! I had the problem also before building ocaml-tree-sitter.

doc/overview.md
```diff
@@ -12,6 +12,22 @@ programming languages. Our goal is to take advantage of these
 grammars and parsers from OCaml, rather than writing and
 maintaining our own.
 
+Definitions
+--
+
+CST: Concrete Syntax Tree. Also known as Parse Tree, it is the
```
aryx, line 18: Nice to define CST and AST.

lang/Makefile.common
```diff
@@ -0,0 +1,68 @@
...
+DUNE_BIN = $(PROJECT_ROOT)/bin
+NODE_BIN = $(PROJECT_ROOT)/node_modules/.bin
+SCRIPTS = $(PROJECT_ROOT)/scripts
+
+# Build the parsers.
+#
+.PHONY: build
+build:
+	$(MAKE) prep
```
aryx, line 14: very nice names. Useful because it can be quickly become confusing with all those code generators involved.

scripts/gen-ocaml
```diff
@@ -1,12 +1,59 @@
...
+# │   └── Main.ml
+# ├── lib
+# │   ├── CST.ml
+# │   ├── dune
+# │   ├── Parse.ml
+# │   ├── parser.c
+# │   ├── scanner.cc
+# │   └── tree_sitter
+# │       └── parser.h
+# └── tree-sitter.opam
```
aryx, line 19: why not tree-sitter-ruby.opam? We will not get conflicts?

lang/Makefile.common
mjambon: The only thing is it depends on git and being inside of a valid git repo, i.e. it doesn't work if you distribute a source tarball without the `.git`.

scripts/gen-ocaml
mjambon: Well-spotted. This was indeed a problem. It's already fixed but I didn't update this comment.

PR #55, opened 2020-06-23
Release generated files + generate boilerplate template

mjambon:
From `lang/`, `make release` will generate the files for the supported languages (currently just `ruby`) and push them to their special repo at https://github.com/returntocorp/ocaml-tree-sitter-lang

There will be a PR for adding it as a submodule to semgrep.

I also added the generation of `Boilerplate.ml`. This is what we get for Ruby: https://github.com/returntocorp/ocaml-tree-sitter-lang/blob/master/ruby/lib/Boilerplate.ml


src/gen/lib/Codegen_boilerplate.ml
```diff
@@ -0,0 +1,120 @@
...
+      sprintf "(%s)"
+        (List.mapi (fun i _ -> sprintf "v%i" (i+1)) l
+         |> String.concat ", ")
+
+let simple_exp x = simple_pat x
+
+let gen_rule_mapper_body body =
+  let arg = simple_exp body in
+  match body with
+  | Symbol name -> [ Line (sprintf "map_%s %s" (trans name) arg) ]
```
aryx, line 51: I would remove the map_, we know the whole file is about a conversion function. See my js_to_generic.ml,
python_to_generic.ml.

src/gen/lib/Codegen_boilerplate.ml
```diff
@@ -0,0 +1,120 @@
...
+      [
+        Line (
+          sprintf "List.map (fun %s -> todo %s) %s"
+            (simple_pat x) (simple_exp x) arg
+        )
+      ]
+  | Choice l ->
+      let cases =
+        List.map (fun (name, body) ->
+          Line (sprintf "| `%s %s -> todo %s"
```
aryx, line 71: I would actually in the body call recursively and then do the todo, for example

| `Cons (a,b,c) -> let a = stmt a in  let b = expr b in let c = token c in todo (a,b,c)

aryx: so we factorize also the boilerplate of calling things recursively

mjambon:
The new version is better: https://github.com/returntocorp/ocaml-tree-sitter-lang/blob/master/ruby/lib/Boilerplate.ml

Major changes:

* I separated the type inlining from their pretty-printing. So we can now transform the grammar to an inlined version of the grammar. The original is still needed to parsing the output of tree-sitter parsers, though. (`Inline` module)
* The generated boilerplate is now more complete.
* I added a `Group` construct to the pretty-printing tree (`Indent` module), which allows collapsing multiple lines into one when the resulting line is not too long.

aryx:
Fantastic! Love all the changes.

PR #56, opened 2020-06-24
Setup

mjambon:
This adds a simple make target to install tree-sitter and ocaml-tree-sitter runtime libs, to be used for semgrep.


mjambon:
I'm merging this to make progress on the semgrep-core side.

PR #57, opened 2020-06-25
Add Dockerfile + CircleCI config

mjambon:
The build uses ocaml 4.10.

* `Dockerfile` works. It serves as a reference and allows local testing.
* The CircleCI setup is a translation of the dockerfile.

CircleCI worked twice and failed once, despite no significant changes. I think we can merge this but there's something fishy that will need to be fixed. See https://app.circleci.com/pipelines/github/returntocorp/ocaml-tree-sitter?branch=ci

The CircleCI run takes between 6 and 10 minutes.

### DIFF
--- a/scripts/install-tree-sitter-lib
+++ b/scripts/install-tree-sitter-lib
@@ -4,8 +4,6 @@
 #
 set -eu
 
-project_root=$(git rev-parse --show-toplevel)
-
 if [[ ! -e tree-sitter ]]; then
   git clone --depth 1 https://github.com/tree-sitter/tree-sitter.git
 fi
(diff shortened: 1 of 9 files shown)


scripts/install-tree-sitter-lib
```diff
@@ -16,4 +14,5 @@ fi
   git clean -dfX
   make
   sudo make install
+  sudo ldconfig /usr/local/lib
```
aryx, line 17: This could be restricting when installing semgrep from source to users.

mjambon: We don't need to install the tree-sitter lib in `/usr/local/lib` (if that's what you mean). I could add a `--prefix` option to our script and call tree-sitter's configure script.

PR #59, opened 2020-07-08
Add a 'configure' step to the build

mjambon:
This script was necessary for building on debian or ubuntu with musl-tools, until I figured this setup wasn't working with C++. We now use alpine for static builds, and the couple flags we need for a static build are set in semgrep-core.

We could get rid of this `configure` script since we don't strictly need it at the moment, but it could be useful for others who might want to build the project in other environments than the ones we use.


aryx:
I guess we can remove the bento checks CI action also.

mjambon:
> BTW did you find the problem with the tree-sitter api.h header file not found in /usr/local/include?

The problems with headers and libraries not being found are not solved. It's the musl-tools setup on debian/ubuntu (using musl-gcc) that's special but now we can deal with it. Other environments (glibc on debian/ubuntu and macos, and musl on alpine) use the usual paths `/usr/local/include` and `/usr/local/lib` without special configuration. We have to call `ldconfig` on debian/ubuntu for the library to be found, and apparently there's no need for the mac equivalent.

PR #64, opened 2020-07-13
comment delete to avoid segfault

aryx:


aryx:
I've managed to avoid some segfaults by calling the parser in a separate process (using Parallel.invoke, Jon Harrop's function), 
but I still get some segfaults because the fork happen at different time and in some situation the memory of the parent
process is close to a Gc and forking at that moment trigger probably a Gc in the child and a segfault. If I comment the ts_delete then there's no segfault at all; There's a giant memory leak, but if we call the parser in a separate process (same fork), at least the leak does not matter.

aryx:
We need to find the right fix for this segfault, but right now that's the best I have to avoid it.

aryx:
This will help https://github.com/returntocorp/enterprise/issues/251

PR #67, opened 2020-07-14
Implement 'children' sequence matching independently from generated code

mjambon:
This large branch reimplements how the ocaml CST is recovered from tree-sitter. The generated files `Parse.ml` are completely different but the type definitions in `CST.ml` are unchanged.

The main highlight is that we can now parse Java and TypeScript in addition to Ruby, thanks to backtracking bugs being now easier to fix.

What changed is the regular expression matching that we use to recover the structure of the `children` fields of the tree-sitter output. We turn such sequence of nodes into an ocaml data structure that reflects repeats, options, sequences, etc. Unlike before, we now operate in two major steps instead of just one:

1. We match generated regular expressions against a sequence of nodes using a plain runtime library (see `Matcher.ml` and `Backtrack_matcher.ml`). If successful, this results in a generic "capture" such as `Repeat [Token "foo"; Token "bar"]` if the input sequence was `["foo"; "bar"]` and the regular expression was `Repeat (Token lowercase_ident)`.
2. Generated code takes care of translating a generic capture into a statically-typed CST, in which each node has a different type.

This doesn't change anything for the caller of a parser. The advantage of this new approach is that we separate concerns. Generated code is big, slow to compile, and hard to debug. Now the tricky part, which consists in regular expression matching, is done separately and we have a whole series of [unit tests that runs fast and tests just that](https://github.com/returntocorp/ocaml-tree-sitter/blob/regexp/src/run/test/Matcher.ml), independently from code generation. The generated code itself is now simpler and easier to review. See Java's [Parse.ml](https://github.com/returntocorp/ocaml-tree-sitter-lang/blob/master/java/lib/Parse.ml) for example.


lang/Makefile
```diff
@@ -4,7 +4,10 @@
 # ocaml-tree-sitter must have been built and installed for this to work.
 #
 
-SUPPORTED_LANGUAGES = ruby
+SUPPORTED_LANGUAGES = \
```
aryx, line 7: woohoo! ok after that we need Go and Python, just to avoid having to fix parsing bugs in pfff for those languages,
especially Go.

aryx:
For example it would be great to explain that  the only difference between Run.Exp.t and Run.Capture.t is
Alt which now has an integer so we "resolved" which branch was taken (which really at some point we should ask tree-sitter folks to include in the tree-sitter output so we don't have to recover it).
Otherwise it's hard to see at first sight the difference, because the types look almost the same, there's just
a Alt of int  in one case.

aryx:
BTW I reread the docs in doc/*.md and those doc are fantastic. I think we just need to extend slightly doc/parsing.md with the separate of concerns discussions and the multi-step approach of Run.Exp, Run.Capture, the generic Matcher and the lang-specific translation from Capture to lang-specific CST.

aryx:
and a small discussion why we use Run.Exp.t type instead of the Choice/Repeat/... in the CST_grammar.ml (I guess because we abstract away a few details there like difference between Token/Pattern/... but that we use those details later when translating  to the lang-specific CST?

PR #73, opened 2020-07-15
Disable warning 42 in generated code.

mjambon:
Fixes https://github.com/returntocorp/ocaml-tree-sitter/issues/72


mjambon:
Mostly because we already have type annotations on the function arguments:
```ocaml
  let children_regexps : (string * Run.exp option) list = [
```

```ocaml
  let trans_variable ((kind, body) : mt) : CST.variable =
```
so I don't think adding an `open` helps the reader.

I also don't think there's value for warning 42 in new code (since ocaml 4.00.1 was released in 2012) but of course existing projects may have it on, and that's fine.

mjambon:
Note that I don't know the minimum ocaml version supported by ocaml-tree-sitter, but I don't think we should bother supporting anything older than (generously) 4 years.

PR #74, opened 2020-07-17
Use pkg-config to locate the tree-sitter library.

mjambon:
Changes include:

* `pkg-config` is relied upon to find the tree-sitter library and headers.
* `brew` is used to install the tree-sitter library on macos (or more exactly on platforms where the `brew` command is available).

The `TREESITTER_LIBDIR` environment variable must be set correctly when compiling files generated by ocaml-tree-sitter. In semgrep, these files are those in `semgrep-core/ocaml-tree-sitter-lang`. `TREESITTER_LIBDIR` must be set with the help of pkg-config prior to calling `dune build` in semgrep, as indicated in the generated `dune` files.

https://github.com/returntocorp/ocaml-tree-sitter-lang/tree/pkgconfig is the generated code that goes with this branch.


mjambon:
This passes the CI tests, which is for linux. It won't work on a system with brew since `brew install tree-sitter` is expected to fail.

scripts/install-tree-sitter-lib
```diff
@@ -9,24 +9,23 @@ error() {
   exit 1
 }
 
-config=config.sh
-[[ -f "$config" ]] || error "Missing config file. Did you run ./configure?"
-. "$config"
-
+echo "Installing tree-sitter library directly from git."
+if [[ ! -e tree-sitter ]]; then
+  git clone --depth 1 https://github.com/tree-sitter/tree-sitter.git
```
aryx, line 14: maybe we want to pin a specific version? just to avoid problems if tree-sitter.git gets incompatible changes.

tree-sitter.opam
```diff
@@ -7,13 +7,15 @@ dev-repo: "git+https://github.com/returntocorp/ocaml-tree-sitter.git"
...
-  ["dune" "build" "-p" name "-j" jobs]
+  ["./configure"]
+  ["make"]
 ]
 
 depends: [
   "alcotest"
   "atdgen"
   "cmdliner"
+  "conf-pkg-config"
```
aryx, line 18: at some point we'll need to write a conf-tree-sitter that does check for tree-sitter and install it if needed, just
like there's a conf-perl, conf-xxx

mjambon: yes!

scripts/install-tree-sitter-lib
mjambon: good idea, I'll pin tree-sitter (after merging this)

PR #77, opened 2020-07-23
Pin tree-sitter to release 0.16.9 (from 5 days ago)

mjambon:


scripts/install-tree-sitter-lib
```diff
@@ -12,7 +12,8 @@ error() {
 
 echo "Installing tree-sitter library directly from git."
 if [[ ! -e tree-sitter ]]; then
-  git clone --depth 1 https://github.com/tree-sitter/tree-sitter.git
+  git clone --depth 1 --branch 0.16.9 \
```
aryx, line 15: Maybe we can put that in a variable before, like PINNED_VERSION

mjambon: I added a command-line option `--version`, defaulting to `0.16.9`.

PR #80, opened 2020-07-25
inline/deinline

mjambon:
I left comments in the code describing what I understand about the behavior of tree-sitter. Here's a summary.

Some rules can be hidden, and can be inlined. Inline implies hidden. And we don't support hidden rules because it makes parsing more complicated for us.

* For hidden rules that are not inlined, we've been simply removing the leading underscore, which introduces new nodes in the grammar. No problem with them.
* Inline rules however can't be turned into normal rules without introducing parsing conflicts. What we've been doing is perform the inlining ourselves by rewriting `grammar.json`. This has been working fine too.

The problem here is that the inlined rules have good, meaningful names which are lost during inlining. This set of changes takes care of keeping the inlined rules around. We then use the factorization algorithm to deinline the types and assign them their original names. So we have type definitions that match the rules in the original grammar before inlining, such as

```ocaml
type anon_choice_get = [
     `Get of Token.t (* "get" *)
   | `Set of Token.t (* "set" *)
   | `Async of Token.t (* "async" *)
 ]
```
which is now
```ocaml
type reserved_identifier = [
     `Get of Token.t (* "get" *)
   | `Set of Token.t (* "set" *)
   | `Async of Token.t (* "async" *)
 ]
```

Variant constructor names no longer start with the type name (due to various complications that come with it), so we'll need to chop off the prefix of variants in the existing code e.g.

```ocaml
 type operator = [
    `Op_DOTDOT of Token.t (* ".." *)
  | `Op_BAR of Token.t (* "|" *)
  | `Op_HAT of Token.t (* "^" *)
  | `Op_AMP of Token.t (* "&" *)
```
is now
```ocaml
type operator = [
    `DOTDOT of Token.t (* ".." *)
  | `BAR of Token.t (* "|" *)
  | `HAT of Token.t (* "^" *)
  | `AMP of Token.t (* "&" *)
```

I'll continue with the fixes on semgrep-core on Monday.

### DIFF
--- a/src/gen/lib/Nice_typedefs.ml
+++ b/src/gen/lib/Nice_typedefs.ml
@@ -4,8 +4,8 @@
 
    1. Undo the inlining of grammar rules due to the 'inline' field
       of the grammar.
-   3. Factor out and assign a name to large nodes that occur multiple times.
-   4. Inline small definitions that are used only once.
+   2. Factor out and assign a name to large nodes that occur multiple times.
+   3. Inline small definitions that are used only once.
 *)
 
 let rearrange_rules grammar =
--- a/src/gen/lib/Nice_typedefs.ml
+++ b/src/gen/lib/Nice_typedefs.ml
@@ -2,11 +2,14 @@
    Improve OCaml type definitions and boilerplate destined for human
    consumption:
 
-   - Inline small definitions that are used only once.
-   - Factor out and name large nodes that occur multiple times.
+   1. Undo the inlining of grammar rules due to the 'inline' field
+      of the grammar.
+   3. Factor out and assign a name to large nodes that occur multiple times.
+   4. Inline small definitions that are used only once.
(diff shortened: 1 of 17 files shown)


src/gen/lib/Nice_typedefs.ml
```diff
@@ -2,11 +2,14 @@
...
-   - Factor out and name large nodes that occur multiple times.
+   1. Undo the inlining of grammar rules due to the 'inline' field
+      of the grammar.
+   2. Factor out and assign a name to large nodes that occur multiple times.
+   3. Inline small definitions that are used only once.
 *)
 
 let rearrange_rules grammar =
   grammar
+  |> Deinline.deinline_rules
```
aryx, line 13: deinline ... inline ...
lol. Sounds like we don't know what we want :)

PR #83, opened 2020-07-29
Dont call sudo when installing tree sitter in homebrew context

brendongo:
Homebrew does not allow calls to sudo when installing semgrep
from source.

### DIFF
--- a/scripts/install-tree-sitter-lib
+++ b/scripts/install-tree-sitter-lib
@@ -56,7 +56,7 @@ fi
 
   # When building from source on homebrew sudo is not allowed
   # but also not needed
-  if [[ -z "${HOMEBREW_SYSTEM}" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/install-tree-sitter-lib
```diff
@@ -53,7 +53,14 @@ fi
...
   make
-  sudo make install
+
+  # When building from source on homebrew sudo is not allowed
+  # but also not needed
+  if [[ -z "${HOMEBREW_SYSTEM}" ]]; then
+    sudo make install
+  else
+    make install
+  fi
```
mjambon, line 63: `[[ -z "${HOMEBREW_SYSTEM+x}" ]]` is the way to go here (works on old versions of bash like the default version on macos).

mjambon:
@brendongo feel free to pull in the latest version of ocaml-tree-sitter into semgrep (commit b340269 is safe).

PR #84, opened 2020-07-30
Import C and C++, plus two necessary fixes in code generation

mjambon:
The first problem was a Choice between multiple rules reduced to Blanks. They were therefore identical and couldn't get a different name, so we now deduplicate the members of a Choice. Whatever the Blanks match, the result doesn't show up in tree-sitter's output... and I don't know where the problem was in the C grammar. I just guessed what the problem was.

The second problem was the "cyclic types abbreviations" as reported by the ocaml compilers, which could be turned off with the `-rectypes` flag. Instead of using `-rectypes`, we detect problematic type definitions (rules) and wrap them in a polymorphic variant (Choice).

Before:
```ocaml
     and new_declarator = (
         Token.t (* "[" *)
       * expression
       * Token.t (* "]" *)
       * new_declarator option
     )
```
After:
```ocaml
     and new_declarator = [
       `Rectype of (
           Token.t (* "[" *)
         * expression
         * Token.t (* "]" *)
         * new_declarator option
       )
     ]
```

There's just one occurrence of this so far, in the C++ grammar.


mjambon:
I also added a C++ example, and I found two problems:

* a bug in ocaml-tree-sitter which I fixed (added the `strip` pass to `CST_grammar_conv`)
* missing support for `and`, `or`, and `not` in the grammar (they may be deprecated or something, need to check)

src/gen/lib/Rectypes.ml
```diff
@@ -0,0 +1,129 @@
...
+
+     and new_declarator = [
+       `Rectype of (
+           Token.t (* "[" *)
+         * expression
+         * Token.t (* "]" *)
+         * new_declarator option
+       )
+     ]
+*)
```
aryx, line 51: fantastic comment! Thx.

src/gen/lib/Rectypes.ml
```diff
@@ -0,0 +1,129 @@
...
+(*
+   Detect and wrap rules so as to not generate cyclic type abbreviations.
+*)
+let prevent_cyclic_type_abbreviations grammar =
+  let rules =
+    List.map (fun rule_group ->
+      rectypes rule_group
+      |> List.map (fun (x, rectype) ->
+        if rectype then
+          { x with body = Choice [("Rectype", x.body)] }
```
aryx, line 123: nice, simple!

PR #90, opened 2020-08-08
Use external grammars extended with semgrep pattern syntax

mjambon:
There's a new repo here: https://github.com/returntocorp/semgrep-grammars
and I left some brief instructions in [doc/adding-a-language.md](https://github.com/returntocorp/ocaml-tree-sitter/blob/semgrep/doc/adding-a-language.md). You can also look at [a CircleCI run](https://app.circleci.com/pipelines/github/returntocorp/ocaml-tree-sitter/156/workflows/00ec5a8a-bd99-4447-be6d-ffeb086e0b88/jobs/160) to see the different steps.

We have even more steps than before, but it's for the best. This avoids copy-pasting, and gives us the opportunity to rename languages or to insert custom steps for this or that language. Now we'll be able to test and extend the original tree-sitter grammars easily thanks to the magic of git submodules.

Current status: only ruby, javascript, and typescript were ported to the new setup. The old setup for the other languages still works.

What I still have to do: review and publish the new generated ocaml code for ruby/javascript/typescript. Will do this on Monday.

After this, there's no rush but we should convert the other languages to the new setup. It shouldn't be hard. An annoyance is going to be the time it takes to fetch the `tree-sitter-*` submodules since they include huge generated `parser.c` files in the git history.


scripts/gen-c
```diff
@@ -14,21 +21,39 @@ EOF
...
 
-test -x "$tree_sitter" || error "missing executable $tree_sitter"
-
 # Generate C source code for the grammar, in two passes:
 #
-# 1. Generate the original grammar.json by executing grammar.js.
-# 2. Simplify grammar.json so it's compatible with ocaml-tree-sitter.
-# 3. Generate the C parser.
+# 1. Import files from semgrep-grammars
+# 1. Simplify grammar.json so it's compatible with ocaml-tree-sitter.
```
colleend, line 27: should this be #2?

mjambon: oops, sorry about that. I rushed this PR as it was Friday night...

mjambon:
ok, I'm merging this to keep things moving. It shouldn't affect production until we release the generated code, which can be done on a per-language basis.

PR #94, opened 2020-08-13
updated README to fill in more info for macOS install. 

colleend:
Also changed uses of cp -r to cp -R as specified in posix. Checked all other scripts, no other uses of cp -r.


mjambon:
This is good. Keep in mind that people don't pay attention to readmes and that readmes, like all text, are not automatically checked so they get obsolete very fast. Anything that can be automated should be automated.

[CircleCI supports macos](https://circleci.com/docs/2.0/hello-world-macos/) so we might as some point set it up to make sure our macos scripts work.

colleend:
that's a very good point. I'll make an issue to integrate a macOS check into CircleCI so that'll be addressed sometime.

PR #106, opened 2020-08-27
"fyi" files

mjambon:
This adds a mechanism to copy arbitrary files into ocaml-tree-sitter-lang at release time. It also adds a much needed `--dry-run` mode.

~~Next, I'll document the process for upgrading a grammar.~~ :heavy_check_mark:


aryx (approved):
Perfect

PR #111, opened 2020-09-22
Add tsx as its own language

mjambon:


aryx:
Does that mean we will have a separate Parse_tsx_tree_sitter.ml or everything is merged with typescript when it comes to ocaml-tree-sitter-lang? (I prefer the later).

mjambon:
No, unfortunately we have two separate grammars, which results in different parsers and different types for the CSTs (due to being auto-generated from the grammar). Thankfully, the types are compatible, so we can convert the typescript CSTs and tsx CSTs to the same type using just type coercion (`:>`). The relevant code in semgrep is https://github.com/returntocorp/semgrep/blob/5fc0b587aaa9569231d9dd0cf335b933445d69fc/semgrep-core/parsing/Parse_typescript_tree_sitter.ml#L2606-L2612

PR #113, opened 2020-09-25
Include hash in names created for anonymous nodes

mjambon:
Implements https://github.com/returntocorp/ocaml-tree-sitter/issues/112

I also added a sorting pass on the list of rules and their dependencies so as to stabilize the order of the definitions (which is the output of topological sort). I updated the generated code in ocaml-tree-sitter-lang in two separate commits to facilitate migration:

1. keep the type names but reorder them according to the new, well-defined order ([bad diff](https://github.com/returntocorp/ocaml-tree-sitter-lang/commit/c5803922c3f1d50a4aede045876f765d265330f5), which doesn't matter).
2. change the type names; their order now mostly remains the same, which results in [good diffs](https://github.com/returntocorp/ocaml-tree-sitter-lang/commit/a887813e00a20f02b3d6337a8b481b35d68fbaab?branch=a887813e00a20f02b3d6337a8b481b35d68fbaab&diff=unified).

I'm now working on the migration in semgrep-core.

### DIFF
--- a/src/gen/lib/Factorize.ml
+++ b/src/gen/lib/Factorize.ml
@@ -155,7 +155,9 @@ let factorize_rules
       && not (Hashtbl.mem node_names node)
       then
         let name =
-          "anon_" ^ Type_name.name_rule_body node |> make_name_unique in
+          sprintf "anon_%s_%s"
+            (Type_name.name_rule_body node) (Type_name.hash_rule_body node)
+          |> make_name_unique in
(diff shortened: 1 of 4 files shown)


mjambon:
`CST.anon_choice_pat_25b90ba_4a37f8c` is a bit too much. I'll do something so we get only one hash in the name.

aryx (approved):

PR #114, opened 2020-10-01
Expose detailed parsing error + highlight snippet in test parsers

mjambon:
A new Error exception is now exposed and designed to be caught an interpreted by the caller. See `Tree_sitter_error.mli`.

Additionally, the error is shown in context when it occurs in a test parser, with color if printing to a terminal:

image

or otherwise as plain text:

```
Error: Line 2, characters 15-17:
function f() {
  thing (); uh oh; thang();
               ^^          
}

node type: ERROR
children: [
  identifier
]
Source code cannot be parsed by tree-sitter.

exit 11
```

It also takes care of truncating strings that are too long so this doesn't result in a disaster when printing error lines from minified files:

image

Takes care of https://github.com/returntocorp/ocaml-tree-sitter/issues/110

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ version: 2.1
 jobs:
   build:
     docker:
-      - image: mjambon/r2c-ocaml:ubuntu
+      - image: returntocorp/ocaml:ubuntu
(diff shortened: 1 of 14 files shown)


aryx (approved):
Great!

PR #117, opened 2020-10-03
merge in stats branch

DrewDennison:


.circleci/config.yml
```diff
@@ -86,8 +89,8 @@ workflows:
   weekly-stats:
     triggers:
       - schedule:
-          # Run at 10:00 every Sunday, UTC.
-          cron: "0 10 * * 3"
+          # Run at 00:00 every day, UTC.
+          cron: "0 0 * * *"
```
mjambon, line 93: You should rename the job to `daily-stats`.

scripts/upload-stat.py
```diff
@@ -0,0 +1,65 @@
+#! /usr/bin/env python3
+#
+# Parse the stat.txt file from lang-stat and upload to the metrics collection endpoint
```
mjambon, line 3: Thank you!

(we could have made the shell script produce json, instead of parsing the `txt` file here)

DrewDennison: yeah I thought about it but figured it wasn't that hard to parse your format and I like your human readable format

PR #120, opened 2020-10-08
Ignore error nodes found in the tree-sitter parse tree

mjambon:
This is an attempt to deal locally with parse errors. Tree-sitter exposes `ERROR` nodes which so far had been causing an exception resulting in failing for the whole input file. This approach removes the error nodes from the tree-sitter parse tree before we try to match its structure against the structure of the grammar. This seems to be working well. This doesn't change the generated CST type.

The only gotcha is that the CST can now omit unparsable parts of the program but appears correct. Maybe we could return the list of errors alongside the CST, and let the user check whether it's empty, if they care.

We'd also have to make sure our parsing stats include these errors.


scripts/parse-examples
```diff
@@ -22,9 +22,23 @@ parser=./parse-"$lang"
 
 # Run the parser from source to ocaml CST
 #
+# Input files are examples/*.src
+# If the test is supposed to fail, the pattern is examples/*.err.src
```
aryx, line 26: FYI I have a small regression parsing library I use for the pfff parser in Common2.regression_testing.
See an example of use here: https://github.com/returntocorp/pfff/blob/develop/lang_java/parsing/test_parsing_java.ml
with that, you parse a directory of test java files and it stores on the disk in /tmp the current score, and if you modifiy the parser and rerun on the same test directory, it will compare with the old result and display regressions.

aryx (approved):
Ok I think I know how to change semgrep to accept errors from the parser. I'll send a PR with the changes in the interface on the semgrep side and you can modify later the ocaml-tree-sitter-lang interfaces to feed me with those errors tokens.

aryx:
| This seems to be working well. This doesn't change the generated CST type.

Did you run stats on your corpus? 

| The only gotcha is that the CST can now omit unparsable parts of the program but appears correct. Maybe we could return the list of errors alongside the CST, and let the user check whether it's empty, if they care.

Yes, we need to change the interface of the Tree_sitter_xxx.Parse to return an AST and a list of errors with their position.
I'll adapt the semgrep side after that to correctly propagate those errors up.

mjambon:
I'm still not finished... but you can look at the new return type of the parse functions. Instead of `CST.foo`, we return `CST.foo option * Tree_sitter_error.t list`.

aryx:
Great. That will also solve by ugly hack around marshalled exceptions if you now use an option and the error in the list.

mjambon:
We still produce a Tree_sitter_error.t exception for some internal errors, i.e. it should not happen just like an Assert_failure or Invalid_argument. I think it's fine to not worry about it because it's not something the user can fix anyway. I might also get rid of it to clarify the interface.

mjambon:
Here are the updated parsing stats, obtained by counting the fraction of lines that don't contain an error. For example, if a file has 100 lines and an error spans 3 lines, the line coverage is now 97%. It used to be 0% since we used to fail for the whole file.

```
=== ocaml-tree-sitter language statistics ===
date: 2020-10-10
git commit: 74eb00e

Language: csharp
Line count: 6240841
Line coverage: 95.463%

Language: go
Line count: 7339138
Line coverage: 99.997%

Language: java
Line count: 4699921
Line coverage: 99.997%

Language: javascript
Line count: 885278
Line coverage: 99.999%

Language: ruby
Line count: 4816201
Line coverage: 98.682%

Language: tsx
Line count: 108523
Line coverage: 99.058%

Language: typescript
Line count: 2235601
Line coverage: 99.905%
```

PR #121, opened 2020-11-03
adding examples, fyi, projects for kotlin

colleend:
Notes:
* changed scripts/lang-stat because `wc -l <X> | cd -d ' ' -f1` command led to empty line count.
* on OSX, grep -P does not exist because the version installed automatically is from FreeBSD and doesn't have the -P flag. Included in the instructions how to bypass this, but we can change grep -P to something else later.

- [x] Added examples for kotlin.
- [x] Added extensions.txt with kotlin extensions.
- [x] Updated adding-a-language doc.
- [x] Created scripts/most-starred-for-language in order to grab most popular github repos for a specific language.
- [x] Added repos to projects.txt
- [x] Updated to new version of semgrep-grammars


scripts/most-starred-for-language.py
```diff
@@ -0,0 +1,33 @@
...
+    s.headers.update({'Accept': 'application/vnd.github.cloak-preview+json'})
+
+    url = 'https://api.github.com/search/repositories?q='+ args.lang + '&s=stars&type=Repositories'
+    r = s.get(url)
+    pretty_json = json.loads(r.text)
+
+    with open('projects.txt', 'w') as f:
+        for repo in pretty_json["items"]:
+            f.write(repo["html_url"]+"\n")
+
```
mjambon, line 33: oh, nice!

mjambon (approved):
awesome

scripts/most-starred-for-language.py
aryx: nice one indeed!

aryx (commented):
Great!

PR #125, opened 2020-11-10
A few projects to test the tree-sitter C and C++ parser

aryx:
test plan:
make stat in the lang/c and lang/cpp directory.
I get 99.5% on git with the C parser and 93.1% on the Cataclysm
project with the C++ parser.

### DIFF
--- a/lang/c/extensions.txt
+++ b/lang/c/extensions.txt
@@ -1,0 +1,3 @@
+# File extensions for the C language
+.c
+
--- a/lang/c/projects.txt
+++ b/lang/c/projects.txt
@@ -1,0 +1,1 @@
+https://github.com/git/git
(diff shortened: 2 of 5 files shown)


aryx:
This will help https://github.com/returntocorp/semgrep/issues/1952

aryx:
There are lots of errors on the git code, but the error-recovery of tree-sitter seems to work pretty well!

aryx:
143 files failing (in lang/c/stat/stat.fail) but still 99.52% line coverage.
I wonder if we undercount errors ... 99.52% seems really high given the number of actual parse errors
there is in the parse-error.log file.

mjambon:
If each error is contained in one line, 99.5% success corresponds to one error every 200 lines. So it's compatible with something like half of the files containing at least one error, which would be 50% success without error recovery.

mjambon (approved):

PR #129, opened 2020-12-15
update to latest ruby

aryx:


aryx:
BTW I've changed the gitmodule URL in semgrep-grammars/ for tree-sitter-ruby to point directly to
tree-sitter/tree-sitter-ruby/ instead of pointing to returntocorp/tree-sitter-ruby. It was adding yet another intermediate
that I don't think is useful in the context of semgrep-grammars and was not in line with what was in updating-a-language.md.

aryx:
This will help https://github.com/returntocorp/semgrep/issues/2231

mjambon:
Having our own forked repo of the grammar is in case we have a branch that hasn't been approved yet by the main tree-sitter-X project. This can be changed as needed.

mjambon (approved):

PR #133, opened 2021-01-06
Update to latest semgrep-grammar with C# and Lua grammar extensions

aryx:


aryx:
Note that I've already did 'make release' to update ocaml-tree-sitter-lang, so I could start writing Lua and C# patterns in semgrep.

PR #135, opened 2021-01-09
Add R support

zythosec:
Adds R language support to ocaml-tree-sitter as part of https://github.com/returntocorp/semgrep/issues/2360


lang/r/projects.txt
```diff
@@ -0,0 +1,30 @@
+https://github.com/binder-examples/r
```
aryx, line 1: I would have expected dlpyr and ggplot2 to be in this list (or those R packages are not in github maybe?)

aryx (approved):

PR #137, opened 2021-01-15
* lang/semgrep-grammars: use latest

aryx:


mjambon:
Hey @Ruin0x11, I moved the file `builtins.lua.src` to `examples.todo/` so as to exclude it from the ocaml-tree-sitter test suite, which otherwise fails. If you think it's a regression rather than a to-do, we need to fix the tree-sitter parser.

mjambon:
I'm in the process releasing the generated C and OCaml files into ocaml-tree-sitter-lang. It's taking its sweet time.

mjambon:
Ran into git-cleanliness issue with tree-sitter-r https://github.com/r-lib/tree-sitter-r/pull/3

mjambon (approved):

mjambon:
The files were released https://github.com/returntocorp/ocaml-tree-sitter-lang/commit/77ed8963bb16332c67a81d2982d76531c3abd3fb. Has updates for rust and lua.

mjambon:
@Ruin0x11 I also moved `read_only_tables.lua.src` to `examples.todo` to avoid errors, like for `builtins.lua.src`.

aryx:
Thx martin!

aryx:
Yes, @Ruin0x11  I think for lua they put this new global_variable rule just in prefix, and it should probably be accepted elsewhere.

PR #139, opened 2021-01-25
json file changed, otherwise git status shows up as unclean

colleend:
When running 

```make release``` 

with the new semgrep-grammars repo, package.json file changes and I can't make a push to ocaml-tree-sitter-lang because the git status is unclean. 

Hopefully with this we're able to push to ocaml-tree-sitter-lang.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,12 @@
+{
+  "dependencies": {
+    "tree-sitter-c-sharp": "file:lang/semgrep-grammars/src/tree-sitter-c-sharp",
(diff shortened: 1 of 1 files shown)


aryx:
No idea why we need this package.json.

aryx (approved):

mjambon:
@colleend I don't know if we need this, since it wasn't a problem before. I may look into it when I get to reorganizing of the tree-sitter-related repos (https://github.com/returntocorp/semgrep/issues/2240).

PR #140, opened 2021-02-06
Include the semgrep-grammars repo instead of using a submodule

mjambon:
* `lang/semgrep-grammars` is now a simple folder, not a submodule anymore
* as a result, we now have a bunch of submodules `lang/semgrep-grammars/src/tree-sitter-*`
* building grammars is now lazy (only happens if inputs are more recent than outputs, like `make` does)
* release generated code for multiple languages at once is now discouraged, and pushing to the main branch of ocaml-tree-sitter-lang is forbidden

I updated the documentation. I should also let the contributors more directly about this change.

This branch should be merged as soon as https://github.com/returntocorp/semgrep/pull/2512 works.


mjambon:
Next, I'll split ocaml-tree-sitter-lang per language so as to make it easy to upgrade or downgrade a language without interfering with the other languages.

aryx (approved):
Great work.

PR #141, opened 2021-02-08
Split ocaml-tree-sitter-lang into one repo per language

mjambon:


doc/adding-a-language.md
```diff
@@ -182,9 +187,10 @@ describing what was expected and what's going on.
 
 ## pfff
 
-Now you need to update pfff, as the generic AST is defined in pfff and
-you will need to specify some details in order to start filling out
-the auto-generated file.
+Pfff defines a list programming languages, some of which have parsers
+in pfff itself. Others are tree-sitter parsers which are otherwise
+independent from pfff. You need to add the new language to the list of
```
aryx, line 192: hmm Lang.ml is now under semgrep-core/Core/, maybe you don't need to modify pfff for that now.

aryx (approved):
Great! This will save us lots of time and avoid lots of problems for contributors.

doc/adding-a-language.md
mjambon: oh, cool.

mjambon: It looks like there's still `file_type.ml` to update in pfff, and the instructions in the pfff readme need to be updated. I'll do that after merging.

PR #157, opened 2021-02-18
Estimate corpus size for parsing stats

mjambon:
This document estimates how many lines of code we should parse to get a good estimate of the parse error rate.

The answer is 10 million, more or less.


aryx (approved):
Fancy!

PR #167, opened 2021-03-02
adding kotlin tests that were in semgrep-core to ocaml-tree-sitter

colleend:
These were existing tests in semgrep-core.


colleend:
This should be in tests/parsing. Ignore pls 😅

PR #168, opened 2021-03-04
adding deep ellipsis operator

colleend:
to Kotlin grammar


aryx (changes requested):
You need to reference deep_ellipsis in expression, like we do for ellipsis, in the choice

aryx (approved):

aryx:
You need a test plan in theory

aryx (approved):
perfect!

PR #171, opened 2021-03-27
Aggregate errors by "class"

mjambon:
It's now part of `make stat`. Two CSV files are generated, showing which errors are more common, according to the number of errors or according to the number of lines affected.

This is what I get with javascript (not many errors):
image

### DIFF
--- a/scripts/aggregate-parse-errors
+++ b/scripts/aggregate-parse-errors
@@ -1,0 +1,44 @@
+#! /usr/bin/env python3
+#
+# Aggregate parse errors into classes so as to give a sense of which errors are
+# more common or serious for a given parser.
+#
+
+import csv
+import json
+import sys
+
+def parse_json_error_log():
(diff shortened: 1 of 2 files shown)


aryx (approved):
This is great. Maybe you could a concrete code example of error in the CSV so has one representative of the error

aryx:
@colleend we could use that to better classify parsing errors in kotlin! Martin improved the parsing script to gather statistics at the parsing error site and group them in clusters!

eatkins:
@mjambon LGTM. Thanks for doing this!

eatkins (approved):

mjambon:
@aryx I'll leave instructions on how to troubleshoot parsing errors to the docs.

colleend:
this is great!! 🙏 🙏 🙏  will definitely be using this for kotlin 😄

PR #172, opened 2021-03-29
Sort output of aggregate-parser-errors

eatkins:
It is easier to identify the high frequency errors if the output is
sorted.

### DIFF
--- a/scripts/aggregate-parse-errors
+++ b/scripts/aggregate-parse-errors
@@ -29,7 +29,7 @@ def parse_json_error_log():
 
 
 def print_csv_clusters(clusters):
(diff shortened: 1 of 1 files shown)


mjambon:
This is fine, but note that `make stat` produces 2 sorted csv files from the original.

mjambon (approved):

eatkins:
Aah, cool. I guess I didn't read through the docs enough and just saw the aggregate-parser-errors script when looking through the git log and tried that first.

eatkins:
Hopefully the subsequent csv sorting is stable so that it's only O(N) to re-sort.

PR #177, opened 2021-04-09
semgrep-rust: Support top-level statements

amietn:
This should make it possible to write Rust semgrep patterns with a list of statements directly.


aryx (approved):
Looks good!

Issue #179, opened 2021-04-11
ocaml-tree-sitter does not generate bindings.c


joseeMDS:
I used `ocaml-tree-sitter ocaml grammar.json` and generated all the ocaml files, but when i tried to build with dune i got `reference not found for octs_create_parser_ocaml"`, after some search i found at [semgrep-ruby repo](https://github.com/returntocorp/semgrep-ruby) that it should also generate a file named bindings.c, maybe i skipped some step?

mjambon:
Yes, `bindings.c` is "generated" from a template that simply inserts the name of the target language. See the script `/scripts/gen-ocaml`. You'll see the result in https://github.com/returntocorp/semgrep-ocaml.

mjambon:
Note that we haven't made efforts to make ocaml-tree-sitter usable outside of semgrep. For now you may have to reimplement partially some of the scripts such as `gen-c` and `gen-ocaml`.

joseeMDS:
Hi @mjambon, thanks for the clarification and the example, the goal is to change the usage of a web-assembly file into an ocaml executable in [melange](https://github.com/melange-re/melange/tree/fork/ocaml-tree). We are using the lib with [esy-tree-sitter](https://github.com/joseeMDS/esy-tree-sitter)  and a resolution to `ocaml-tree-sitter`.

I would be happy in contributing to add `ocaml-tree-sitter` to opam

mjambon:
@joseeMDS we'd love to have ocaml-tree-sitter in opam, go for it. I created issue #181 for discussing this.

Issue #178, opened 2021-04-10
Simplify-grammar dont stop


joseeMDS:
I am trying to use ocaml-tree-sitter with ocaml, i did the following steps:

```bash
tree-sitter generate
ocaml-tree-sitter ocaml src/grammar.json
```
and i got `Error: Aliases are not suported`
I found the simplify-grammar command and tried it at the `grammar.json` folder but has passed 5 hours and there is no results, is this the supposed behavior?

joseeMDS:
Thanks to @anmonteiro that showed me how to do it, i didnt know that `simplify-grammar` was expecting receive something from stdin, tried: `cat grammar.json | simplify-grammar` and worked

ievans:
Ah glad you got some help @joseeMDS !

Sorry if docs were unclear, would happily accept a pr if you or @mjambon have a clarification in mind!

joseeMDS:
I would be happy contributing, i will open a PR adding some docs ASAP

mjambon:
@joseeMDS we have some docs in https://github.com/returntocorp/ocaml-tree-sitter/tree/master/doc. In particular, [Code generation pipeline](https://github.com/returntocorp/ocaml-tree-sitter/blob/master/doc/code-generation-pipeline.md) explains a bit what's going on.

We're not using the tree-sitter-ocaml grammar in semgrep at this point, but we could add it to `/lang` in ocaml-tree-sitter anyway. There's a script for this, which hopefully will work (see instructions in [How to add support for a new language](https://github.com/returntocorp/ocaml-tree-sitter/blob/master/doc/adding-a-language.md)). Then the generated code would be published to its own repo, which could be plugged into an ocaml project as a git submodule [like we do in semgrep](https://github.com/returntocorp/semgrep/tree/develop/semgrep-core/src/tree-sitter-lang).

This complicated setup is designed to accommodate grammar extensions used in semgrep patterns, such as `...`. If we don't do anything, then it will be identical to the original tree-sitter-ocaml grammar. Let me see if I can set this up, it will be useful to us as well.

mjambon:
I just added ocaml to the list of tree-sitter languages in ocaml-tree-sitter. The generated code is at https://github.com/returntocorp/semgrep-ocaml. This should build without difficulties if you add this folder into an existing dune project.

PR #180, opened 2021-04-11
Add tree-sitter-ocaml to languages

mjambon:
This will show us how well the tree-sitter parser for ocaml works. It also helped answering questions from @joseeMDS who wants to use the ocaml interface to the parser (see #178, #179).

Stats:
```
Line count: 3729862
Line coverage: 99.133%
```
I haven't looked into the errors. There's a chance of .ml files using camlp4 syntax extensions. Those should be excluded.

The generated code is at https://github.com/returntocorp/semgrep-ocaml


aryx (approved):

aryx:
This can help https://github.com/returntocorp/semgrep/issues/2885

PR #184, opened 2021-04-12
Update tree-sitter-rust submodule to point to latest commit

amietn:
It was previously pointing to an oudated Rust grammar version


aryx:
Hmm, actually this is not up-to-date with the latest tree-sitter-rust.
The submodule listed for tree-sitter-rust in ocaml-tree-sitter is https://github.com/Ruin0x11/tree-sitter-rust because
Ruin0x11 started the work to add support for Rust in semgrep, but now we should go back to use the
official tree-sitter-rust at https://github.com/tree-sitter/tree-sitter-rust

aryx (changes requested):
Update the submodule to point to the official tree-sitter-rust, not Ruin0x11's fork.

aryx (approved):

PR #189, opened 2021-04-14
Add version info to "FYI" files released alongside generated code

mjambon:
See e.g. https://github.com/returntocorp/semgrep-typescript/blob/main/fyi/versions

I hope it will make it easier to see which version of grammar.js was used to derive the code.


mjambon:
I also added a file containing the tree-sitter version.

aryx (approved):

PR #194, opened 2021-04-19
Remove homebrew special case

brendongo:
Had separate installation steps for when in homebrew environment.
Moved special casing to semgrep repo instead of this repo


mjambon:
Nice. Thank you @brendongo . Sorry for the delay, I was working on a talk.

PR #196, opened 2021-04-20
Add support for new 'precedences' field in tree-sitter grammars + upgrade javascript grammar

mjambon:


mjambon:
hmm, looks like typescript depends on our local version of the javascript grammar and needs to be upgraded as well.

mjambon:
Now both javascript and typescript grammars are set to use the same version of tree-sitter-javascript.

mjambon:
I pushed the generated parsers for javascript, typescript, and tsx. If CI passes on this branch, we're good to go.

mjambon:
CI failed again, this time with a conflict in the tsx grammar. Works fine on my machine. I don't know what's going on. Pausing this effort for now.

mjambon:
Things seemed to work for me because some dependency hadn't been refreshed. I still haven't figured all the issues with `npm install`; not sure I ever will. 

Anyway, some changes needed to be made on the typescript grammar, which are now merged.

PR #197, opened 2021-04-24
Add conf-tree-sitter to tree-sitter.opam

joseeMDS:
With this we should be able to guarantee that tree-sitter is installed and even install it with `opam depext`

### DIFF
--- a/tree-sitter.opam
+++ b/tree-sitter.opam
@@ -25,6 +25,7 @@ depends: [
   "ppx_sexp_conv"
   "sexplib"
   "tsort" {>= "2.0"}
+  "conf-tree-sitter"
 ]
 
 synopsis: "Code generator for parsing the output of tree-sitter parsers"


aryx:
But we need a specufic version of treesitter >= 0.19

joseeMDS:
the package with `opam depext` will install the latest version since it will install from the OS package manager, but itself will only check if tree-sitter is installed, so if the user has an older tree-sitter installed, it would be a problem

mjambon:
I'm not clear on the best practices for managing the version of libtree-sitter. I'm guessing we could use one of the following solutions:
* a version constraint on `conf-tree-sitter` such as `"conf-tree-sitter" {= "0.19"}` where the version of the `conf-tree-sitter` package matches the version of libtree-sitter.
* a dedicated `conf-tree-sitter-0-19` package name (with its own, not very important package version)

I suggest asking at https://github.com/tree-sitter/tree-sitter (use the "Discussions" board or file an issue) for guidance regarding versioning, compatibility and the best way of achieving what we want (including possibly having multiple versions of the tree-sitter library installed on the same system).

mjambon:
I'm splitting ocaml-tree-sitter into 3 repos (#204) to help separate concerns. I'm leaving this open so we can continue the conversation. Code changes for the ocaml/opam stuff will now be made in the new repo ocaml-tree-sitter-core.

mjambon:
@joseeMDS There are now two new repos of interest to you: ocaml-tree-sitter-core and ocaml-tree-sitter-languages. See the main readme.

PR #198, opened 2021-05-06
Add subcommands

joseeMDS:
This is the progress i have made so far, i still have to add `safe_run` equivalent for each command, but my main question right now is if running a shell script in OCaml is a good idea, maybe we could convert `gen_c` and `gen_ocaml` into `gen/lib/Generate.ml`? and then just call `Generate.main` like in the simplify command.  what do you think?


mjambon:
Looks great so far!

mjambon:
> my main question right now is if running a shell script in OCaml is a good idea, maybe we could convert `gen_c` and `gen_ocaml` into `gen/lib/Generate.ml`?

I wrote `gen-c` and `gen-ocaml` in bash because it was quick for me to set up. I think the main arguments for switching to ocaml are:
* It's more appealing to contributors, who are familiar with ocaml but may not be comfortable with bash and all its gotchas.
* It doesn't add complexity since we already have dune set up.

The cons are:
* I don't like the syntax of non-shell languages (python, ocaml etc.) for dealing with external commands and IO redirections.
* As a consequence I'm not familiar with writing this kind of code in ocaml.

Of course, if you're interested in doing this translation work for `gen-c` and `gen-ocaml`, that would be great. Do you have a plan in mind, like using [shcaml](https://tov.github.io/shcaml/doc/) maybe? From the docs, it seems like a totally new world with a lot of things to (re)learn rather than mapping 1:1 with shell scripts. For example, how would we translate `mkdir -p folder; cp file folder`? Should we call `cp` or use an ocaml function for copying files? If there was a good collection of how-to guides that might help.

joseeMDS:
Hmmm, `Sys` and maybe the `Unix` module should be enough, but there are a few things that can be annoying, like `Sys.mkdir` is only available in OCaml >= 4.12, and `Unix.mkdir` would not work on windows, so maybe the best option is use the scripts, now my question is what we should do to find them, maybe rely on some env var?

mjambon:
Most of the `Unix` works on Windows. I don't think our scripts do anything that wouldn't work on Windows: creating, editing, inspecting files and folders; running commands in the foreground.

If we keep the scripts as is, we should be able to install them into opam's `bin` or other folder managed by opam which is in the opam user's PATH. How about we just rename the scripts by sticking an `ocaml-tree-sitter-` prefix and install them alongside other executables in opam?

joseeMDS:
Yeah, this should be enough, i will rename the scripts and add the safe commands functions

mjambon:
There's a conflict on `gen-c`.

It looks good, thank you. Let's merge this if it passes the tests!

mjambon (approved):

mjambon:
oops, the CI tests didn't run, I think because the branch comes from an external repo.

mjambon:
I think I enabled CircleCI for external pull requests now.

mjambon:
Sorry about the mess, @joseeMDS. I reverted master, and pushed your original branch to this repo, under the name `add-subcommands` so you can recover it if it was deleted from your repo.

I made a new PR just to see what the tests tell us, and so far I saw the following:
* need to run `pre-commit run -a` to reformat the files (see instructions in the main readme)
* some scripts are still calling `gen-c` (old name)

If you take the branch and make another PR, you should be able to see the CircleCI build tests running. It should look like this:
image

PR #202, opened 2021-05-12
Add subcommands

joseeMDS:
Same as #198 but now with CI running to check the tests, and fix the scripts names


joseeMDS:
oh, there are some conflicts with the scripts name and ocaml-tree-sitter binary, not sure why

joseeMDS:
Okay, looks like when naming the scripts like `ocaml-tree-sitter-*`, the binary will try to execute it as a subcommand, so how we should name the scripts?

mjambon:
The problem is that [the call to ocaml-tree-sitter](https://github.com/joseeMDS/ocaml-tree-sitter/blob/f6cd3b0b018c20d2e59cf2a6b5c2e63358e135a8/scripts/ocaml-tree-sitter-gen-ocaml#L140-L144) now requires a subcommand. It thinks that `seq` (the first argument) is the subcommand.

I suggest creating a short subcommand for the what used to be the only mode of operation of ocaml-tree-sitter. I.e. `ocaml-tree-sitter [ARGS]` would become `ocaml-tree-sitter gen [ARGS]`. The command-line `--help` should then list both `gen` and `simplify` as subcommands.

joseeMDS:
The test are passing, but the semgrep action failed, looks like there is no authentication token @mjambon, anyway i will add a commit improving help page and i think that's all

mjambon:
> the semgrep action failed

I set this up yesterday for this repo, and it's not working for you because you're on a forked repo and need an access token. I think you could set one up if you wanted to. We might want to improve the error message produced by semgrep.

Anyway, don't worry about semgrep failing.

mjambon:
> The tests are passing

:tada: awesome, let's merge!

mjambon (approved):

PR #206, opened 2021-05-20
Extend javascript and typescript grammar with semgrep constructs

mjambon:
- [x] import tests from pfff
- [ ] extend syntax for javascript
- [ ] extend syntax for typescript
- [ ] add/review tests for each semgrep construct


aryx (approved):
Nice. Nice flexibility of tree-sitter here, to easily extend an existing grammar with
additional constructs.

aryx:
I mean, we knew it was flexible, but this is the first time we extensively extend an original grammar with all semgrep constructs.

mjambon:
Adding a note about how to maintain this, then merging. For now, these features will remain unused by semgrep.

PR #207, opened 2021-05-24
removing Scala

aryx:
Using tree-sitter-scala to parse Scala code proved to be harder than
we thought given the parsing ambiguities of the Scala language.
In the end we opted to mostly porting the existing recursive-descent
parser in the Scala compiler to OCaml.

test plan:
make


mjambon:
I removed scala from the list of languages in the makefile.

mjambon (approved):

aryx:
oops, thanks!

PR #210, opened 2021-06-10
Go deeper into the workaround to support 'token()'

mjambon:
We initially hit an error with `token(prec.dynamic(1, 'as'))` in the tree-sitter-hack grammar, and then we (original author and myself) didn't find another way of writing the grammar. The problem is that `token(...)` sometimes produces a node and sometimes doesn't, and the previous workaround consisting in creating a named rule for `token(...)` would make the precedence annotation ineffective. 

I had to [figure out in which cases a token node is missing](https://github.com/tree-sitter/tree-sitter/issues/1156) and handle these cases accordingly. This results in 3 cases, described and tested in the script `tests/inline-token/check-test-output`.

The file `Name_pattern.ml` that was renamed `Missing_node.ml` for clarity on its purpose. This is where we have the logic for working around missing nodes [due to patterns (regexps)](https://github.com/tree-sitter/tree-sitter/issues/1151) or due to the `token()` construct.

Conclusion: a `token(prec.dynamic(/[a-z]+/))` will produce no node in the original CST and a node of type unit in the typed CST, with no way to recover the token unless it has its own rule in the original grammar. Other combinations are handled fine.


aryx (commented):
Nice comments, nice tests.

PR #219, opened 2021-06-24
Update security policy for releasing generated files

mjambon:


mjambon:
I'm merging this to avoid dead links from the generated repos like https://github.com/returntocorp/semgrep-c, but please let me know what you think, @mschwager .

PR #229, opened 2021-07-07
Add tree-sitter-sqlite

mjambon:


.gitmodules
```diff
@@ -59,3 +59,6 @@
 [submodule "lang/semgrep-grammars/src/tree-sitter-dockerfile"]
 	path = lang/semgrep-grammars/src/tree-sitter-dockerfile
 	url = https://github.com/camdencheek/tree-sitter-dockerfile.git
+[submodule "lang/semgrep-grammars/src/tree-sitter-sqlite"]
+	path = lang/semgrep-grammars/src/tree-sitter-sqlite
+	url = https://github.com/returntocorp/tree-sitter-sqlite.git
```
aryx, line 64: why not on the original tree-sitter-sqlite? Why our cloned version?

aryx (approved):

.gitmodules
mjambon: I did that because I had two pending pull requests (https://github.com/dhcmrlchtdj/tree-sitter-sqlite/pulls?q=is%3Apr+is%3Aclosed). The author merged them already, so I'll switch back to the official repo.

PR #230, opened 2021-07-08
adding regex project and deleting kotlin lang

colleend:
* deleting jetbrains/kotlin because the project uses self-made test patterns that have not been seen in other projects. The self-created test patterns usually do not parse well with the current ocaml-tree-sitter kotlin grammar.


aryx (approved):
Good.

PR #236, opened 2021-07-16
Warn against Blank nodes during test-lang

mjambon:
I [added this to the documentation](https://github.com/returntocorp/ocaml-tree-sitter-semgrep/blob/main/doc/updating-a-grammar.md#testing) earlier and figured this check would be better automated.

The affected languages are currently:
* C (c)
* C++ (cpp)
* go
* hcl

cc @Sjord who suffered from this before with the C# grammar.

### DIFF
--- a/lang/test-lang
+++ b/lang/test-lang
@@ -72,6 +72,27 @@ if [[ "$lang" = typescript ]]; then
   sublangs="typescript tsx"
 fi
 
+check_for_blanks() {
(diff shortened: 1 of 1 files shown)


aryx (approved):

aryx:
Maybe in the error message we could point to a PR that fixed one of those issues, so one can imitate easily.

PR #239, opened 2021-08-03
add extension: .hcl

dhutty-numo:
Several applications other than Terraform use HCL (both other HashiCorp apps and non-hashi, e.g. Terragrunt)


mjambon:
Great, thanks!

PR #240, opened 2021-08-09
Adding tree-sitter-php

aryx:


aryx:
cc @Sjord

lang/php/projects.txt
```diff
@@ -0,0 +1,3 @@
+# Git URLs of publicly-accessible projects to be used for parsing stats,
+# one per line.
+#
```
aryx, line 3: @mjambon could you polulate this? I tried the script but it has asking for a github API key.

aryx:
This will help https://github.com/returntocorp/semgrep/issues/3576

lang/php/projects.txt
aryx: actually I just copy-pasted the one we had in semgrep/parsing-stats/php/

mjambon: yeah you have to generate an api key for yourself, preferably with just the read permission so you can copy-paste it in your shell without worrying.

PR #241, opened 2021-08-10
Move documentation to semgrep.dev

emjin:
Move documentation in ocaml-tree-sitter-semgrep/doc to semgrep.dev (semgrep-docs repo). See returntocorp/semgrep-docs#245


mjambon (approved):

mjambon:
Note that now there's also [ocaml-tree-sitter-languages](https://github.com/returntocorp/ocaml-tree-sitter-languages) which is a permanent fork of ocaml-tree-sitter-semgrep. This is the community repo for folks using ocaml-tree-sitter without semgrep. No need to do anything, it already has warnings, see https://github.com/returntocorp/ocaml-tree-sitter-languages/blob/main/doc/adding-a-language.md

PR #244, opened 2021-08-10
[Hack] Extend statements and expressions with ellipsis

frankeld:
- Updates tree-sitter-hack to the latest version
- Adds ellipsis support in statements, expressions, member declarations, and parameter (testing included as well)

Build should match https://github.com/frankeld/semgrep-hack/tree/release-6f0bf4c

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-hack/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-hack/grammar.js
@@ -13,23 +13,58 @@ const base_grammar = require('tree-sitter-hacklang/grammar');
 module.exports = grammar(base_grammar, {
   name: 'hack',
 
-  conflicts: ($, previous) => previous.concat([
-  ]),
+  conflicts: ($, previous) => previous.concat([]),
 
-  /*
-     Support for semgrep ellipsis ('...') and metavariables ('$FOO'),
-     if they're not already part of the base grammar.
-  */
   rules: {
-  /*
-    semgrep_ellipsis: $ => '...',
+    /*
+      Support for semgrep ellipsis ('...')
+    */
+
+    // For now, we give precedence to native ellipsis (aka variadic modifiers).
+    ellipsis: ($) => prec(-1, '...'),
+    deep_ellipsis: ($) => prec(-1, seq('<...', $._expression, '...>')),
+
+    // By using an empty statement, we leverage a statement type that is strictly only used within
+    // $._statement and nowhere else.
+    empty_statement: ($, previous) => {
+      return choice(
+        previous,
+        // For conflict with `ellipsis;`
+        // we want it to be read as an expression with a semicolon
+        // instead of a statement
+        prec(-1, $.ellipsis) // statement ellipsis
+      );
+    },
 
     _expression: ($, previous) => {
-      return choice(
-        $.semgrep_ellipsis,
-        ...previous.members
+      return choice(previous, $.ellipsis, $.deep_ellipsis);
+    },
+
+    member_declarations: ($) => {
+      return seq(
+        '{',
+        choice.rep(
+          // Copied from existing grammar
+          alias($._class_const_declaration, $.const_declaration),
+          $.method_declaration,
+          $.property_declaration,
+          $.type_const_declaration,
+          $.trait_use_clause,
+          $.require_implements_clause,
+          $.require_extends_clause,
+          $.xhp_attribute_declaration,
+          $.xhp_children_declaration,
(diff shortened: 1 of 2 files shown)


frankeld:
Also, I manually ran regression tests as the automatic ones aren't compatible as of now. See https://github.com/slackhq/tree-sitter-hack/pull/27.

lang/semgrep-grammars/src/semgrep-hack/grammar.js
```diff
@@ -13,23 +13,58 @@ const base_grammar = require('tree-sitter-hacklang/grammar');
...
+        previous,
+        // For conflict with `ellipsis;`
+        // we want it to be read as an expression with a semicolon
+        // instead of a statement
+        prec(-1, $.ellipsis) // statement ellipsis
+      );
+    },
+
+    _expression: ($, previous) => {
+      return choice(previous, $.ellipsis, $.deep_ellipsis);
```
frankeld, line 40: Why do semgrep extension files return the values? When I tried just setting, it worked the same as far as I can tell. Is it to just differentiate new rules vs changed rules to the reader?

mjambon: I'm not sure what you mean. Are you just being confused by the different ways to define anonymous functions?

```js
(x) => x
```
is the same as
```js
(x) => { return x; }
```
which is itself the same as
```js
function(x) { return x; }
```

When extending a `choice()` we can use two ways:

* The inline way, which works only for `choice()` (e.g. not `seq()`). It really extends the original list of choices:
```js
_expression: ($, previous) => {
  return choice(...previous.members, $.ellipsis, $.deep_ellipsis);
}
```

* The simpler way:
```js
_expression: ($, previous) => {
  return choice(previous, $.ellipsis, $.deep_ellipsis);
}
```
which is more compactly written as
```js
_expression: ($, previous) => choice(previous, $.ellipsis, $.deep_ellipsis)
```

The simpler way results in extra nesting, and you'll see that in the ocaml code. I don't know if there's a strong reason to use one or the other. I imitated [what was done for tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/common/define-grammar.js), which heavily extends tree-sitter-javascript.

mjambon (approved):
looks great to me!

frankeld:
Thanks for the review @mjambon ! Just pushed up more stuff for metavariables. Other PRs were updated as well. Should save you from doing too many semgrep-hack builds :)

mjambon (approved):

lang/semgrep-grammars/src/semgrep-hack/grammar.js
frankeld: This answers my question. I was just confused about the code style choices. In the [original file](https://github.com/returntocorp/ocaml-tree-sitter-semgrep/blob/926b452e6c3e819a27b8477254ae31937ba1ca16/lang/semgrep-grammars/src/semgrep-hack/grammar.js#L25-L30), one rule is written with the `(x) => x` syntax and one rule is written with the `(x) => { return x; }`. I didn't know if there was an implicit preference in certain cases.

mjambon: Yeah, don't assume I was clear with this syntax myself when I wrote or copied that code :stuck_out_tongue: I learn a lot by explaining these things and writing docs.

mjambon: I improved the template in this PR: https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/245

Also, I turned the comments above into documentation: https://semgrep.dev/docs/contributing/adding-a-language/#extending-the-original-grammar-with-semgrep-syntax

mjambon:
I'm not sure what you meant by "Other PRs were updated as well" but I made a release. [CST.ml](https://github.com/returntocorp/semgrep-hack/blob/main/lib/CST.ml) looks about right.

frankeld:
> I'm not sure what you meant by "Other PRs were updated as well" but I made a release. [CST.ml](https://github.com/returntocorp/semgrep-hack/blob/main/lib/CST.ml) looks about right.

The pluralization of "PR" here might have been a bit misleading. I was just saying that https://github.com/returntocorp/semgrep/pull/3704 was also updated to support the new CST and the branch on my fork of semgrep-hack has an updated reference build. 

✅ https://github.com/returntocorp/semgrep/pull/3704 now points to the new semgrep-hack release.

PR #252, opened 2021-09-16
Use back the original fcwd/tree-sitter-kotlin repo

aryx:
I get 71% parsing stats with it instead of 69% with
our fork (which contains a few improvements).
I think we should restart from scratch and make proper
PR to this original repo.

test plan:
```
$ cd lang ./test-lang kotlin
...
OK
$ cd kotlin
$ ~/github/ocaml-tree-sitter-semgrep/scripts/lang-stat kotlin projects.txt extensions.txt
...
71%
```

### Security

- [ ] Change has security implications (if so, ping the security team)


colleend (approved):
nice!

PR #256, opened 2021-10-04
Use latest tree-sitter-kotlin

aryx:
### Security

- [ ] Change has security implications (if so, ping the security team)


colleend (approved):
👍 👍

PR #259, opened 2021-10-15
Add a bash grammar extension/hack to support assignment to metavariables

mjambon:
### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-bash/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-bash/grammar.js
@@ -21,6 +21,24 @@ module.exports = grammar(base_grammar, {
       previous
     ),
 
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-bash/grammar.js
```diff
@@ -84,6 +102,8 @@ module.exports = grammar(base_grammar, {
     ),
 
     semgrep_metavariable: $ => /\$[A-Z_][A-Z_0-9]*/,
+    semgrep_metavar_eq: $ => /\$[A-Z_][A-Z_0-9]*=/,
```
aryx, line 105: what if there is space before the '='?

aryx (approved):

PR #260, opened 2021-10-25
Add support for the Elixir language

fpmosley:
This request is to add support for the Elixir language in `semgrep`. Recently, an officially supported [Elixir grammar for tree-sitter](https://github.com/elixir-lang/tree-sitter-elixir) was released by the Elixir team, and that grammar was used to add support for Elixir in `semgrep`.

After generating the parser, parsing statistics were collected using `make stat` on the Top 30 Elixir projects in Github sorted by most starred. The [elixir-lang](https://github.com/elixir-lang/elixir) repo was also included in those projects. The results were:
```
Line count: 705469
Line coverage: 98.153%
```

[Elixir at PagerDuty](https://www.pagerduty.com/blog/elixir-at-pagerduty/) has been a thing since 2018 when we first started migrating some existing services to, and writing new services in, the Elixir language. With that being said, I was able to collect additional stats by running the generated parser across the top 10 active PagerDuty internal/private repos.

`make stat-priv` results
```
Line count: 143449
Line coverage: 99.994%
```

The generated parser works for `*.ex` and `*.exs` extensions.

I was also wrote a simple "Hello World" script in Elixir for the `test/ok` directory and the grammar passes `make test`. Additional tests for ok and fail should be written. 

### Security

- [x ] Change has no security implications (otherwise, ping the security team)


aryx:
This was helping: https://github.com/returntocorp/semgrep/issues/3698

PR #266, opened 2021-11-12
use latest kotlin

aryx:
### Security


aryx:
Hmm, we have regressions on semgrep-rules with that update. Let me check first.

aryx:
@mjambon looks like the CI does not use the semgrep-specific scanner.c I provided. It's weird because it works locally.

colleend (approved):
lgtm! wierd that CI doesn't use scanner.c

mjambon:
@pad ok, what you did to provide your own `scanner.c` makes sense. I don't know if there's a better solution. It's the first time we do this.

mjambon (approved):

PR #267, opened 2021-11-13
Simplify semgrep extensions for dockerfile

mjambon:
### Security


aryx (approved):
You should maybe add some tests in corpus/semgrep.txt

PR #269, opened 2021-11-18
[C#] support for method chaining

aryx:
This will help https://github.com/returntocorp/semgrep/issues/4300

test plan:
test file included
./test-lang c-sharp

### Security


aryx:
cc @Sjord

PR #271, opened 2021-12-01
Add support for solidity, a smart contract language

aryx:
test plan:
./test-lang solidity

### Security


colleend (approved):
very exciting!! looking forward to solidity support 😁

PR #272, opened 2021-12-02
Add stats for solidity with projects.txt

aryx:
test plan:
```
cd lang
./test-lang solidity
cd solidity
make stat
```
=> 99.208% on 324550 LOC, not bad!

### Security


colleend:
wow, very good stats!

PR #275, opened 2022-01-04
Switch back to official tree-sitter-rust

aryx:
We were using Ruin0x11 fork, but if we want to go forward
we need to go back to the official tree-sitter-rust.
This removes the 3 commits Ruin0x11 added on top of an
old version of tree-sitter-rust. This PR switch
back to the common ancestor. Next PR will use the latest
tree-sitter-rust

test plan:
./test-lang rust
cd rust
make stat

still at 99.3%

### Security


aryx:
This may help https://github.com/returntocorp/semgrep/issues/4444 at some point

mjambon (approved):

PR #277, opened 2022-01-05
Use latest tree-sitter-rust

aryx:
### Security


aryx:
The parsing stats actually go down to 99.1% vs 99.3% before, but anyway we need to start
to improve things from the latest tree-sitter-rust.

PR #278, opened 2022-01-06
Adding tree-sitter-python

aryx:
test plan:
./test-lang python
cd python
make stat

### Security


aryx:
This might help https://github.com/returntocorp/semgrep/issues/4039 at some point

aryx:
Line count: 5061118
Line coverage: 99.925%

PR #280, opened 2022-01-12
More semgrep extensions for dockerfile

mjambon:
Most metavariables are already covered by `$.expansion`, which is a shell-like variable expansion and string concatenation supported by dockerfile.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
@@ -62,11 +62,23 @@ module.exports = grammar(base_grammar, {
       )
     ),
 
+    // override
     label_pair: ($, previous) => choice(
       $.semgrep_ellipsis,
-      previous
+      seq(
+        field(
+          "key",
+          choice(
+            $.semgrep_metavariable,
+            alias(/[-a-zA-Z0-9\._]+/, $.unquoted_string)
+          )
+        ),
+        token.immediate("="),
+        field("value", choice($.double_quoted_string, $.unquoted_string))
+      )
     ),
 
+    // TODO: find a way to support metavariables as env keys.
     env_pair: ($, previous) => choice(
       $.semgrep_ellipsis,
       previous
@@ -108,7 +120,7 @@ module.exports = grammar(base_grammar, {
       FROM extras:$CODE_VERSION       # metavariable (pattern only)
       FROM extras:${CODE_VERSION}     # parameter from ARG
 
-      mv: (tricky like ENV)
+      mv: done
       LABEL $NAME=$VAL
 
       e: done
--- a/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
@@ -35,12 +35,25 @@ module.exports = grammar(base_grammar, {
       $.semgrep_metavariable
     ),
 
-    // TODO: support metavariables and ellipses in a bunch of places
+    /*
+      Metavariable syntax vs. ARG expansions:
+
+      - in a semgrep pattern, $FOO is always a metavariable.
+      - in a semgrep pattern, ${FOO} is always an arg expansion.
+      - in a semgrep pattern, $FOO in a position where a metavariable
+        is not allowed results in an error.
+
+     TODO: support metavariables and ellipses in a bunch of places
+    */
     semgrep_metavariable: $ => /\$[A-Z_][A-Z_0-9]*/,
     semgrep_ellipsis: $ => '...',
 
     /*
       TODO: more syntax ellipsis (e) or metavariables (mv):
+
+      mv:
+      FROM extras:$CODE_VERSION       # metavariable (pattern only)
+      FROM extras:${CODE_VERSION}     # parameter from ARG
 
       e, mv:
       LABEL multi.label1="value1" multi.label2="value2" other="value3"
--- a/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
@@ -35,6 +35,38 @@ module.exports = grammar(base_grammar, {
       $.semgrep_metavariable
     ),
 
+    // redefinition
+    arg_instruction: ($, previous) => seq(
+      alias(/[aA][rR][gG]/, "ARG"),
+      field(
+        "name",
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
```diff
@@ -35,52 +35,147 @@ module.exports = grammar(base_grammar, {
...
+    */
     semgrep_metavariable: $ => /\$[A-Z_][A-Z_0-9]*/,
     semgrep_ellipsis: $ => '...',
 
     /*
       TODO: more syntax ellipsis (e) or metavariables (mv):
 
-      e, mv:
-      LABEL multi.label1="value1" multi.label2="value2" other="value3"
+      mv: done
```
aryx, line 119: Are those mv and e defined somewhere? After a few seconds I guessed it meant metavariables and ellipsis
but could be nice to write it somewhere.

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
```diff
@@ -65,3 +65,380 @@ RUN ["a", $X, "b"]
...
+=============================================
+
+FROM --platform=$PLATFORM debian
+ADD --chown=$USER_GROUP src dst
+
+---
+
+(source_file
+  (from_instruction
+    (param)
```
aryx, line 80: this --platform=$PLATFORM is reduced just to a 'param'?

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
```diff
@@ -65,3 +65,380 @@ RUN ["a", $X, "b"]
...
+FROM $NAME:$TAG@$DIGEST
+FROM ${NAME}:${TAG}@${DIGEST}
+
+---
+
+(source_file
+  (from_instruction
+    (image_spec
+      (image_name
+        (expansion
```
aryx, line 101: I thought $NAME here would be not considered a variable expansion but a metavariable.
Is it because there's a : after?

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
```diff
@@ -65,3 +65,380 @@ RUN ["a", $X, "b"]
...
+          (variable)))
+      (image_tag
+        (expansion
+          (variable)))
+      (image_digest
+        (expansion
+          (variable)))))
+  (from_instruction
```
aryx, line 109: So looks like there is no difference between use $XXX or ${XXX}, both result in (expansion (variable)).
Is it intented?

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
```diff
@@ -65,3 +65,380 @@ RUN ["a", $X, "b"]
...
+
+ADD ... dst
+COPY ... dst
+VOLUME ...
+
+---
+
+(source_file
+  (add_instruction
+    (path)
```
aryx, line 183: Ellipsis is reduced to 'path'? Is it intended too? Shouldn't we have a special (ellipsis) construct in the CST?

lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
mjambon: yes, it's on the line above. :wink:

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
mjambon: yes. The details are not shown due to hidden rules (something like `_param_key`), and we can't unhide them without breaking the tests inherited from the original tree-sitter-dockerfile.

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
mjambon: `$NAME` and `${NAME}` are docker argument expansions. We'll inspect the token when converting the CST into a semgrep pattern to determine if it's a metavariable.

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
mjambon: yes. I didn't write that, it's dockerfile syntax.

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
mjambon: `...` is a valid path, so again we need to inspect the token when translating a pattern and convert it to an ellipsis if its value is `...`. When parsing a dockerfile target, `...` will remain a string.

lang/semgrep-grammars/src/semgrep-dockerfile/grammar.js
mjambon: I'll remove these comments when I'm done with the translation to the semgrep AST.

aryx: oops ...

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
aryx: This could be a useful comment to put in the test file.

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
aryx: This could be a useful comment to put in the test file.

aryx (approved):

lang/semgrep-grammars/src/semgrep-dockerfile/test/corpus/semgrep.txt
mjambon: yes. I don't know how to put comments in tree-sitter test files. I'd rather avoid using the target language's comment syntax. I asked here: https://github.com/tree-sitter/tree-sitter/discussions/1586

PR #282, opened 2022-02-08
Update to latest tree-sitter-solidity and add ... in contract

aryx:
This will help https://github.com/returntocorp/semgrep/issues/4587

test plan:
./test-lang solidity

### Security


colleend (approved):
nice!!

PR #284, opened 2022-02-12
Fix broken link for grammar upgrade docs

nmote:
### Security

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ $ ./test-lang kotlin
 ```
 
 For details, see [How to upgrade the grammar for a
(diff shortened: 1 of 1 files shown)


aryx (approved):

nmote:
Looks like the build is broken. I'm seeing the same failure on main

aryx:
@mjambon can you unblock Nat?
Nat is a contractor working for us and for now fixing bugs here and there.

mjambon:
The problem is with the version of cmdliner. We had this problem on other repos and we haven't fixed it here. (This is Hack Week at r2c so I'm not going to fix it this week)

PR #289, opened 2022-03-10
use latest tree-sitter-c-sharp

aryx:
test plan:
./test-lang c-sharp
cd c-sharp
make stat

### Security


aryx:
This will help PA-945

PR #290, opened 2022-03-11
[HTML] support metavar for tags

aryx:
This overrides a very old PR:
https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/234

test plan:
./test-lang html

### Security


aryx:
This will help https://github.com/returntocorp/semgrep/issues/4078

PR #291, opened 2022-03-11
Fix: Allow 'CMD ...' to match any CMD instruction

ajbt200128:
Part of fix for  returntocorp/semgrep#4770

Will need an additional commit when returntocorp/tree-sitter-dockerfile#1 is merged and released
### Security

- [ X] Change has no security implications (otherwise, ping the security team)


mjambon:
superseded by https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/293 (doesn't require change in the original grammar)

PR #295, opened 2022-03-23
Update to latest Go

aryx:
Also fixed errors from ocaml-tree-sitter about
a Blank node.
See
https://github.com/tree-sitter/tree-sitter-go/commit/e9e693780c647c46c37336af9a4b44d343a526f6

test plan:
./test-lang go

### Security

### DIFF
--- a/lang/test-lang
+++ b/lang/test-lang
@@ -66,7 +66,7 @@ fi
 
 check_for_blanks() {
   cst_ml=ocaml-src/lib/CST.ml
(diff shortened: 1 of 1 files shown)


aryx:
reach 99.971% (before was 99.908% but actually was not even building because of Blank nodes errors).

aryx:
This will help https://github.com/returntocorp/semgrep/issues/4823

PR #302, opened 2022-05-10
feat(swift): Parse semgrep_ellipsis

nmote:
This introduces a `semgrep_ellipsis` rule that is a valid expression. It
conflicts with the existing unbounded range operator. This conflict is
resolved as described in the comment.

Test plan:

Existing tests ensure that unbounded ranges are parsed as they used to
be. The added tests make that clear, and also demonstrate a case where
an ellipsis is now parsed. That example used to be a parse error.

### Security


lang/semgrep-grammars/src/semgrep-swift/grammar.js
```diff
@@ -10,21 +10,24 @@ module.exports = grammar(base_grammar, {
...
      if they're not already part of the base grammar.
   */
   rules: {
-  /*
-    semgrep_ellipsis: $ => '...',
+    // Swift has an unbounded range operator which is also three dots. When we
+    // are parsing a pattern, we will convert that into an ellipsis after it's
+    // parsed, but let's use low precedence here so that normally we only
+    // produce this node where the unbounded range operator cannot exist.
+    semgrep_ellipsis: $ => prec.dynamic(-1337, $._three_dot_operator_custom),
```
mjambon, line 25: good comment! 💯

mjambon (approved):

PR #305, opened 2022-05-17
feat(swift): Parse ellipsis in all positions needed for experimental

nmote:
I also added the deep ellipsis here.

Test plan:

Automated tests

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-swift/test/corpus/semgrep.txt
+++ b/lang/semgrep-grammars/src/semgrep-swift/test/corpus/semgrep.txt
@@ -8,7 +8,7 @@ Ellipsis
 
 (source_file
   (additive_expression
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-swift/test/corpus/semgrep.txt
```diff
@@ -23,3 +23,70 @@ Unbounded range operator
   (additive_expression
     (tuple_expression (fully_open_range))
     (integer_literal)))
+
+================================================================================
+Param Ellipsis
```
aryx, line 28: Those are calls arguments, not parameters :)

nmote: :facepalm:

PR #308, opened 2022-05-26
feat(rust): ellipsis for method and field chains

kopecs:
### Security


mmcqd (approved):
I believe you should an add entry to `semgrep-grammars/src/semgrep-rust/test/corpus/semgrep.txt` describing the addition to the grammar, [like we have in C#](https://github.com/returntocorp/ocaml-tree-sitter-semgrep/blob/main/lang/semgrep-grammars/src/semgrep-c-sharp/test/corpus/semgrep.txt#L223). Other than that looks great!

lang/semgrep-grammars/src/semgrep-rust/grammar.js
```diff
@@ -51,10 +51,20 @@ module.exports = grammar(standard_grammar, {
...
         $.ellipsis,
-        $.deep_ellipsis
+        $.deep_ellipsis,
+        $.member_access_ellipsis_expression,
       );
     },
 
+    // TODO: have to use 13 instead of PREC.field because the Rust grammar
+    // doesn't export precedences. Should we use something like
+    // https://github.com/jhnns/rewire to access this directly?
```
aryx, line 61: dunno about this, but if it works ... why not.
The project page of rewire mentions it's about unit test, so not sure it applies here, because it's monkey patching
for the main code, not unit test, but if it works for that case too, whatever!

lang/semgrep-grammars/src/semgrep-rust/test/corpus/semgrep.txt
```diff
@@ -91,6 +91,57 @@ impl Foo {
       (function_item (identifier) (parameters)
         (block (let_declaration (identifier) (deep_ellipsis (integer_literal))))))))
 
+=====================================
+Ellipsis for method / field chains
```
aryx, line 95: There's no ellipsis for method chain here ...
Here is the semgrep.txt for C#:
```
=====================================
Ellipsis in method chain
=====================================
$X = $O.foo(). ... .bar();

---
(compilation_unit
  (global_statement
   (expression_statement
     (assignment_expression
       (identifier)
       (assignment_operator)
       (invocation_expression
         (member_access_expression
           (member_access_ellipsis_expression
              (invocation_expression
                (member_access_expression (identifier) (identifier))
               (argument_list))
              (ellipsis))
          (identifier))
         (argument_list))))))
```

kopecs: 🤦‍♂️ I'll make the change to that and update it. I guess I just had "method chain" in my head and forgot the ellipsis part.

lang/semgrep-grammars/src/semgrep-rust/grammar.js
kopecs: I saw some similar comments on our extensions elsewhere. A better solution would perhaps be to see if we can just get the upstream grammars to export the precedence levels?

lang/semgrep-grammars/src/semgrep-rust/test/corpus/semgrep.txt
kopecs: #310

PR #313, opened 2022-06-06
update to latest tree-sitter-ts and tree-sitter-js

aryx:
test plan:
./test-lang typescript
./test-lang javascript

### Security


mjambon:
If I read the log correctly, there's now one test that passes but was expected to fail:
```
XPASS: test/xfail/pfff/parsing/big/minimized.min.js
```

mjambon (approved):

PR #316, opened 2022-06-14
R: allow metavariable in semgrep-r

aryx:
test plan:
./test-lang r

### Security


aryx:
This will help https://github.com/returntocorp/semgrep/issues/2360

PR #323, opened 2022-06-29
Julia: Adding a few projects in projects.txt

aryx:
test plan:
./test-lang julia
cd julia
make stat

### Security


aryx:
We reach 83% when discarding julia's compiler repository which usually is not representative of what developers
write (usually it contains lots of test cases).

83% is not bad, but it's also not great.

aviks:
> 83% is not bad, but it's also not great.

Yes, I know we need to improve the tree-sitter parser significantly. We're on it, but I did want to make parallel progress on semgrep.

PR #324, opened 2022-06-29
Migrate from `semgrep-action` to `semgrep ci`

kopecs:
The former is sunsetting and we should move to the native functionality. See also https://github.com/returntocorp/semgrep-action/blob/develop/README.md

### Security


aryx:
This had the advantage though of seeing problems customers had!

aryx (approved):

PR #327, opened 2022-07-28
chore(swift): ellipsis as class members

brandonspark:
This PR adds a test and functionality for supporting ellipses in the body of a class.

### Security


lang/semgrep-grammars/src/semgrep-swift/test/corpus/semgrep.txt
```diff
@@ -90,3 +90,29 @@ Deep Ellipsis
...
+(source_file (class_declaration (type_identifier) (class_body (semgrep_ellipsis))))
+
+================================================================================
+Class Ellipsis Sandwiched 
+================================================================================
+
+class ClassName {
+  ...
+  var x = 2
+  ...
```
nmote, line 113: I don't think this got parsed the way you expect. These two lines are a single statement. Maybe try adding a semicolon after the variable declaration? tbh I don't remember exactly what Swift allows here, but there ought to be a way to tweak this so that the second `...` is also parsed as a semgrep ellipsis.

PR #328, opened 2022-07-28
chore(swift): swift stats

brandonspark:
Added some projects and the Swift extension.

### Security


aryx (approved):
So, what are the current stats?

brandonspark:
something pretty good, like 99.5%. though, this is all tree-sitter's success rate, and we had nothing to do with it, right?

aryx:
yep, we have nothing to do with it :)

PR #329, opened 2022-07-29
Kotlin: support ellipsis for method chaining for semgrep

aryx:
This will help https://github.com/returntocorp/semgrep/issues/5819

test plan:
test file included

### Security


colleend:
yay!

PR #334, opened 2022-08-17
upgraded python tree sitter grammar

brandonspark:
**What:**
We don't support Python match constructs in our parser, because tree-sitter is out of date.

**Why:**
https://github.com/returntocorp/semgrep/issues/5885

### Security


aryx (approved):

aryx:
well we don't use yet tree-sitter-python to parse python code in semgrep, but I guess it's one step in the right direction.

PR #339, opened 2022-08-31
java: Parse semgrep extensions

nmote:
Currently we use pfff to parse Java patterns. Unfortunately, due to an
issue I do not fully understand, it is unable to parse the pattern
`public Foo() { }`, which is a standalone constructor with a privacy
modifier. It looks like the grammar should allow this, and in fact it
does allow patterns like `Foo() { }`. There is already a hack in place
to address an LR(1) conflict, and I suspect that it is not working as
desired. Rather than debug the issue with the pfff parser, I think it
makes sense to switch to using the tree sitter parser as a fallback for
pattern parsing like I recently did for JS/TS.

This does the minimum to support the basic Semgrep pattern extensions
(ellipsis in a few positions and tests verifying that metavariables
parse), as well as the constructor case that does not work correctly in
the pfff parser.

Helps https://github.com/returntocorp/semgrep/issues/5558

Test plan: Automated tests, plus testing of integration with semgrep
(already done, but PR not yet up).

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-java/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-java/grammar.js
@@ -10,6 +10,12 @@ module.exports = grammar(base_grammar, {
   name: 'java',
 
   conflicts: ($, previous) => previous.concat([
+    // Conflict due to the addition of semgrep ellipsis to both. Can occur in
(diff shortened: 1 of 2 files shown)


aryx (approved):
Great!

PR #342, opened 2022-09-07
Update latest tree-sitter java

bmahe:
### Security


aryx:
What are the statistics with this new version?
(cd lang; ./test-lang java; cd java; make stat)

aryx (approved):

PR #343, opened 2022-09-12
chore: Update tree-sitter-swift and work around custom operator issue

nmote:
Notably, this pulls in:
* Using the external scanner to handle custom operators: https://github.com/alex-pinkus/tree-sitter-swift/commit/c17937810b59388e818ffb493a45fe424b85698c
* No longer using the external scanner to handle "...": https://github.com/alex-pinkus/tree-sitter-swift/commit/f58deb71df91bcee6d650774dbd136a7493ca20f

# Custom operator change:

Unfortunately, `...>` is a valid custom operator in Swift, and it can occur as part of an expression. So, for the pattern `<... 5 ...>`, it gets lexed as `<...`, followed by the expression `5`, followed by a `custom_operator` token. This doesn't fit the deep ellipsis rule, so it results in a parse error.

A proper fix would be to extend the scanner so that `...>` is its own token, and then handle it separately in the grammar. This would be complicated, and would likely present significant maintainability challenges because the way custom scanners are structured does not invite extensibility. We would probably have to maintain a fork of the scanner.

Instead, I propose this hack. Unfortunately, it will prevent people from writing patterns like `<... 5 .+. 1 ...>` because the custom operator in the middle will get mistaken for the end of the deep ellipsis. However, this can easily be worked around with parentheses, and it's probably going to come up rarely, if ever.

When incorporating this into Semgrep, I plan to check to ensure that the text of the custom operator is `...>` and fail if not. This way, we won't have backwards-compatibility issues if we do this better in the future due to people inadvertently writing `..>` or the like.

# "..." change:

This change involved making a few tweaks to our grammar extensions, but nothing really major. The big thing that this changes is that now the following program is parsed as an identifier followed by, as a separate statement, a fully open range:

```
foo
...
```

When parsing a Semgrep pattern, we convert a fully open range into a Semgrep ellipsis. So, when used as a pattern, this will mean "the statement `foo` followed by any number of statements`, which aligns with user expectations. Previously, it was parsed as a single expression statement: an open-ended range.

Test plan: Automated tests

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-swift/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-swift/grammar.js
@@ -11,6 +11,7 @@ module.exports = grammar(base_grammar, {
 
   conflicts: ($, previous) => previous.concat([
     [$._three_dot_operator, $.semgrep_expression_ellipsis],
+    [$._three_dot_operator, $.semgrep_expression_ellipsis, $.semgrep_ellipsis],
   ]),
 
   /*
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-swift/grammar.js
```diff
@@ -22,11 +23,14 @@ module.exports = grammar(base_grammar, {
...
-    semgrep_expression_ellipsis: $ => prec.dynamic(-1337, $._three_dot_operator_custom),
+    semgrep_expression_ellipsis: $ => prec.dynamic(-1337, "..."),
 
+    // Use "..." here and in semgrep_expression_ellipsis above, rather than
+    // `$._three_dot_operator`, so that the conflicts list is easier to
+    // understand.
     semgrep_ellipsis: $ => "...",
 
-    semgrep_deep_ellipsis: $ => seq("<...", $._expression, "...>"),
+    semgrep_deep_ellipsis: $ => seq("<...", $._expression, $.custom_operator),
```
aryx, line 33: this would maybe benefit from a comment, about the fact that ...> is parsed as a custom operator
(and the hack/problem it comes with).

nmote: Good point. I should have added a comment inline here in addition to explaining it in the commit summary. I'll follow up.

PR #344, opened 2022-09-12
java: Parse semgrep ellipsis as a statement

nmote:
This should cover pattern parsing for nearly all patterns. There are still a few pattern features for GA languages that won't work with this parser (dots method chaining, deep expr operator, etc.). We'll add those later.

Test plan: Automated tests

### Security


bmahe (approved):

lang/semgrep-grammars/src/semgrep-java/grammar.js
```diff
@@ -16,6 +16,20 @@ module.exports = grammar(base_grammar, {
     // are already several conflicts in the original grammar to deal with this
     // case.
     [$.primary_expression, $.formal_parameter],
+
+    // We need to add semgrep_ellipsis to the statement rule, so that the
```
aryx, line 20: great comment! Thx!

PR #352, opened 2022-10-03
chore: update csharp

brandonspark:
### Security


nmote:
Looks like the CI job is getting killed while generating the C# parser... you are probably going to need @mjambon's help to figure this out.

PR #353, opened 2022-10-14
Set up Apex

mjambon:
The generated code for case-insensitive keywords is [ugly](https://github.com/returntocorp/semgrep-apex/blob/32eca0baa1f54f41372609ed7a4f03862fdd3804/lib/CST.ml#L15):
```
type pat_ba503d3 = Token.t (* pattern [eE][xX][cC][lL][uU][dD][eE][sS] *)
```
I'm considering improving the code gen to instead produce:
```
type excludes = Token.t (* pattern [eE][xX][cC][lL][uU][dD][eE][sS] *)
```

Other than that, it works. The parsing stats on the top 20 repos from GitHub are:
```
Line count: 901986
Line coverage: 99.940%
```
which is excellent.

### Security


aryx (approved):
Excellent!

PR #362, opened 2022-11-02
chore: update kotlin

brandonspark:
### Security


nmote (approved):
Looks like tree-sitter-kotlin didn't change the scanner this time, so this should be safe.

PR #378, opened 2022-12-31
augment julia with ellipsis

brandonspark:
## What:
I just added ellipsis into the Semgrep grammar, along with some tests

"interpolation expression" is the category that expressions like `$X` fall into, as those are genuine Julia language constructs. We'll just convert it to a metavariable during Generic conversion.

### Security


aryx (approved):
perfect!
Love the corpus examples.

PR #380, opened 2023-01-06
Update core

mjambon:
Update core. This changes what's generated for `Boilerplate.ml`. See:
* changes in code generation: https://github.com/returntocorp/ocaml-tree-sitter-core/pull/58
* sample diff for Apex: https://github.com/returntocorp/semgrep-apex/commit/f694acd85b352d6359b889b7b20a8d8b16d4b173

### Security


aryx (approved):

aryx:
I'm curious how this will change the process to write the next Parse_xxx_tree_sitter.ml.
The advantage of todo raising exceptions is that it returns an 'a, so it's easy to gradually
migrate code to the generic AST.

mjambon:
Let me know if it makes things more difficult but I'm hopeful. At worst, we can always replace all the `R.XXX` constructors by `todo`.

PR #381, opened 2023-01-18
Solidity: fix ellipsis in contract inheritance

Raz0r:
### Description
Right now if a contract inherits several other contracts, it is not possible to find a specific one, e.g. this rule will not match:

```solidity
contract A is B, C, D {}
```
```yaml
- pattern: contract $A is C { ... }
```
Ellipsis will give an error:
```yaml
- pattern: contract $A is ..., C, ... { ... }
```
This PR resolves parsing error and hopefully allows to match specific inherited contracts using ellipsis.
### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-solidity/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-solidity/grammar.js
@@ -90,6 +90,13 @@ module.exports = grammar(base_grammar, {
             );
         },
 
+        inheritance_specifier: ($, previous) => {
+            return choice(
+                previous,
+                $.ellipsis
+            );
+        },
+
       // The actual ellipsis rules
         deep_ellipsis: $ => seq(
             '<...', $._expression, '...>'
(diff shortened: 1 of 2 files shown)


aryx:
yes I agree, the first pattern should also match, but that's something for Generic_vs_generic or Parse_solidity_tree_sitter.ml

aryx (approved):

aryx:
I've updated semgrep-solidity: https://github.com/returntocorp/semgrep-solidity/commit/948a024c37565b7d4d1a4ecf834238cdce5bcb1d

Raz0r:
Thank you @aryx! The only thing left is to update submodule "semgrep-solidity" in https://github.com/returntocorp/semgrep/tree/develop/languages/tree-sitter-lang

aryx:
Fixed by https://github.com/returntocorp/semgrep/pull/6936
Should make it for the next release (1.6.0).

PR #383, opened 2023-01-29
Jsonnet: add metavar and ellipsis for jsonnet

aryx:
I need to make a PR for the tree-sitter-jsonnet patch.

test plan:
./test-lang jsonnet

### Security


aryx:
@mjambon I had to patch the original tree-sitter-jsonnet with a hack to be able to override 'id'.
Let me know if we can do it in another way.
(the maintainer of tree-sitter-jsonnet already accepted a few my PRs so hopefully he will accept this hack too).

PR #385, opened 2023-02-03
XML: add support for xml by tweaking semgrep-html

aryx:
test plan:
./test-lang html
./html/parse-html ~/pom.xml
which works

### Security


aryx:
This will help https://github.com/returntocorp/semgrep/issues/5939

PR #388, opened 2023-02-20
fix circleCI, do not use xlarge resource and skip c# and kotlin

aryx:
This seems to not be part of our plan anymore.
We get
```
Resource class docker for xlarge is not available for your project, or
is not a valid resource class. This message will often appear if the
pricing plan for this project does not support docker use.
```
error in circleCI dashboard

test plan:
wait for circleCI to kick in and hope for green check

### Security

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
     # - medium+: 6 GB
     # - large: 8 GB
     # - xlarge: 16 GB
-    resource_class: xlarge
+    #resource_class: xlarge
 
     working_directory: ~/ocaml-tree-sitter
     steps:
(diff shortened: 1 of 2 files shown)


mjambon (approved):

mjambon:
I don't know why our CircleCI plan keeps being downgraded to the free tier. I added @DrewDennison as a reviewer.

What we need is take the time to fix the memory/CPU blow-up in `tree-sitter generate`.

aryx:
yes, I also asked drew to put us back the performance plan. I don't think it's very expensive.

aryx:
In any case, even with commenting those languages the job is not finishing, so we need a performance plan.

PR #390, opened 2023-02-22
Add semgrep constructs to ruby

emjin:
### Security


lang/semgrep-grammars/src/semgrep-ruby/grammar.js
```diff
@@ -12,6 +12,14 @@ module.exports = grammar(standard_grammar, {
   rules: {
     /* Ruby global variables start with a '$' ('global_variable'). */
 
+    semgrep_dots: $ => '...',
```
aryx, line 15: they're called semgrep_ellipsis in all of our other grammars, maybe worth keeping it consistent.

aryx (approved):

PR #392, opened 2023-02-25
Add comment explaining what I've tried with ruby pattern parsing

emjin:
### Security


emjin:
@aryx I gave up on adding semgrep constructs. If you're curious why see my changes

aryx (approved):

PR #398, opened 2023-03-15
chore: update ruby tree-sitter parser and tests

brandonspark:
This PR just updates the ruby `tree-sitter` parser. One of the xfail tests now passes, as well, so I moved that to `ok`.

### Security


aryx (approved):
Thanks!

PR #400, opened 2023-03-17
feat!: add Standard ML language

brandonspark:
Yes, I have an agenda.

This is step 1 in integrating Semgrep into my summer class' curriculum, so that I can use Semgrep for pedagogical purposes.

We don't necessarily need to advertise SML as "experimental" once it's in (unless you think it would benefit), I just want it for my own use.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-sml/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-sml/grammar.js
@@ -1,0 +1,46 @@
+/*
+  semgrep-sml
+
+  Extends the standard sml grammar with semgrep pattern constructs.
+*/
+
+const base_grammar = require('tree-sitter-sml/grammar');
+
+module.exports = grammar(base_grammar, {
+  name: 'sml',
+
+  conflicts: ($, previous) => previous.concat([
+    [$.mrule, $._atpat]
+  ]),
+
+  /*
+     Support for semgrep ellipsis ('...') and metavariables ('$FOO'),
+     if they're not already part of the base grammar.
+  */
+  rules: {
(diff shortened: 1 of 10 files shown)


lang/semgrep-grammars/src/semgrep-sml/grammar.js
```diff
@@ -0,0 +1,46 @@
...
+  ]),
+
+  /*
+     Support for semgrep ellipsis ('...') and metavariables ('$FOO'),
+     if they're not already part of the base grammar.
+  */
+  rules: {
+    semgrep_ellipsis: $ => '...',
+
+    // As it turns out, $x parses as the application of `$` to an identifier.
```
aryx, line 23: yes but how is parsed $ X
with a space in between?
Probably better to make a special token for $XXX with token.immediate. There should be no conflict.

lang/sml/projects.txt
```diff
@@ -0,0 +1,18 @@
...
+https://github.com/standardml/twelf
+https://github.com/Quantomatic/quantomatic
+https://github.com/RedPRL/sml-redprl
+https://github.com/urweb/urweb
+
+# This one's for you, Sam.
+https://github.com/shwestrick/smlfmt
+
+# :^)
+https://github.com/brandonspark/mulligan
```
aryx, line 18: haha :)

aryx (approved):
Nice.
What are currently the parsing stats? 100%?

How do you plan to translate to the generic AST? Going through ast_ml.ml intermediate first?
or directly to AST_generic.ml?
I would vote for going to ast_ml.ml first, so we can also improve ast_ml.ml for better dogfooding of OCaml in semgrep.

PR #402, opened 2023-03-17
improve sml, change projects

brandonspark:
This PR adds in better metavariables, no longer relying on translating the applications themselves. I couldn't use `token.immediate` though, as it would make it no longer work, over `token`.

One of the projects was bringing our parse rate way down. Don't trust N64-related programs written in SML, I guess: https://github.com/shygoo/n64sym/blob/master/src/builtin_signatures.sig

Our parse rate is now 99.928%. Sorry, I forgot to say earlier what it was.

### Security


lang/semgrep-grammars/src/semgrep-sml/grammar.js
```diff
@@ -23,6 +23,13 @@ module.exports = grammar(base_grammar, {
     // As it turns out, $x parses as the application of `$` to an identifier.
```
aryx, line 23: update the comment then?

aryx (approved):

PR #404, opened 2023-03-22
feat(haskell): add metavariable support

brandonspark:
This PR adds metavariables to the Haskell grammar. Ellipses were too hard, see my comments.

### Security


lang/semgrep-grammars/src/semgrep-haskell/test/corpus/semgrep.txt
```diff
@@ -0,0 +1,191 @@
+================================================================================
+Simple metavariable
+================================================================================
+
+$X = 2
+
+--------------------------------------------------------------------------------
+
+(haskell
+  (function
```
aryx, line 10: weird that this is parsed as a function.

aryx (approved):

aryx:
nice!

PR #407, opened 2023-04-20
feat(kotlin): add named ellipses

brandonspark:
This PR adds named ellipses to Kotlin.

### Security


brandonspark:
@aryx

PR #408, opened 2023-05-03
chore: fork `tree-sitter-kotlin`, and update tree-sitter kotlin

brandonspark:
There has been a lack of responsiveness from the maintainers of https://github.com/fwcd/tree-sitter-kotlin, and we need to make progress. So let's fork the `tree-sitter-kotlin` repository. We can go back later if we need to.

(I don't know if this is actually the right way to do it)

@aryx @bmahe 

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-kotlin/src/scanner.c
+++ b/lang/semgrep-grammars/src/semgrep-kotlin/src/scanner.c
@@ -213,11 +213,11 @@ bool scan_multiline_comment(TSLexer *lexer) {
         if (after_star) {
           after_star = false;
           nesting_depth--;
-          // if (nesting_depth == 0) {
+          if (nesting_depth == 0) {
             lexer->result_symbol = MULTILINE_COMMENT;
             mark_end(lexer);
             return true;
-          // }
+          }
         } else {
           after_star = false;
           if (lexer->lookahead == '*') {
--- a/lang/semgrep-grammars/src/semgrep-kotlin/test/corpus/semgrep.txt
+++ b/lang/semgrep-grammars/src/semgrep-kotlin/test/corpus/semgrep.txt
@@ -16,10 +16,11 @@ class $CLASS {
     (class_body
       (function_declaration
         (simple_identifier)
-        (parameter
-          (simple_identifier)
-          (user_type
-            (type_identifier)))
+        (function_value_parameters
+          (parameter
+            (simple_identifier)
+            (user_type
+              (type_identifier))))
         (function_body
           (statements
             (call_expression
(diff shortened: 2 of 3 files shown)


lang/semgrep-grammars/src/semgrep-kotlin/src/scanner.c
```diff
@@ -213,11 +213,11 @@ bool scan_multiline_comment(TSLexer *lexer) {
         if (after_star) {
           after_star = false;
           nesting_depth--;
-          // if (nesting_depth == 0) {
+          if (nesting_depth == 0) {
```
aryx, line 216: what are those changes?

lang/semgrep-grammars/src/semgrep-kotlin/test/corpus/semgrep.txt
```diff
@@ -16,10 +16,11 @@ class $CLASS {
     (class_body
       (function_declaration
         (simple_identifier)
-        (parameter
-          (simple_identifier)
-          (user_type
-            (type_identifier)))
+        (function_value_parameters
```
aryx, line 19: why this extra node?

aryx (approved):
Let's do it! Let's fork it.

lang/semgrep-grammars/src/semgrep-kotlin/test/corpus/semgrep.txt
brandonspark: it was added via https://github.com/returntocorp/tree-sitter-kotlin/commit/f4eb099b34d1bf13b0e1dbc3e250199f605cdeb1, but not updated within our `semgrep-kotlin` test suite because we hadn't updated to that version of `tree-sitter-kotlin` yet

lang/semgrep-grammars/src/semgrep-kotlin/src/scanner.c
brandonspark: these are ported changes from the `scanner.c` in `tree-sitter-kotlin`. we have to manually update it because we keep a corresponding version, but with some minor semgrep-specific changes. these correspond to the scanner changes i made in https://github.com/returntocorp/tree-sitter-kotlin/commit/e3bbe6080df359a60c91f4d2645e3510cc6343fe to the normal `tree-sitter-kotlin`

aryx: Ah ok. I hope the semgrep-specific stuff are clearly tagged with a comment.

aryx (approved):

PR #411, opened 2023-05-08
[Cairo 1.0] Improve Semgrep extension grammar

Frostweeds:
### Context

The current grammar does not support some simple construct like `let $X = 42;` due to the fact that the metavariable is in the wrong production. Additionally, they were minor issues in the original tree-sitter grammar that prevented some semgrep construct to be properly interpreted.

This PR fix these issues and allow richer Semgrep patterns to be used on Cairo 1.0

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-cairo/test/corpus/semgrep.txt
+++ b/lang/semgrep-grammars/src/semgrep-cairo/test/corpus/semgrep.txt
@@ -1,0 +1,101 @@
+===================
+Semgrep Constructs
+===================
+
+#[derive(..., Unsafe, ...)]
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-cairo/test/corpus/semgrep.txt
```diff
@@ -0,0 +1,101 @@
+===================
+Semgrep Constructs
+===================
```
aryx, line 3: we usually try to have smaller tests and each test testing just one thing.

aryx (approved):
perfect

PR #412, opened 2023-05-09
fix: use semgrep branch for tree-sitter-kotlin

brandonspark:
after https://github.com/returntocorp/tree-sitter-kotlin/pull/7, we'd like to use the `semgrep` branch for `tree-sitter-kotlin`, which is where we will merge future changes to `tree-sitter-kotlin`.

### Security


aryx (approved):
Maybe add a comment somewhere saying we must use the semgrep branch of our tree-sitter-kotlin fork!

PR #413, opened 2023-05-16
swift: Support ellipsis in navigation suffix

ihji:
returntocorp/semgrep#7664

### Security


aryx:
You can use ./release swift to update the semgrep-swift repo

PR #414, opened 2023-05-17
swift: Support named ellipsis

ihji:
returntocorp/semgrep#7666

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-swift/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-swift/grammar.js
@@ -30,6 +30,8 @@ module.exports = grammar(base_grammar, {
     // understand.
     semgrep_ellipsis: $ => "...",
 
+    semgrep_ellipsis_metavar : $ => /\$\.\.\.[a-zA-Z_][a-zA-Z_0-9]*/,
+
     /*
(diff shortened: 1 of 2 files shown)


nmote:
Looks good, but these are called ellipsis metavariables. Could you update the names?

ihji:
I used the same convention here: https://github.com/returntocorp/ocaml-tree-sitter-semgrep/blob/main/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js#L31

but I appreciate your suggestion to align with the official document naming. It would be better to follow the naming conventions specified in the official documentation.: https://semgrep.dev/docs/writing-rules/pattern-syntax/?gclid=CjwKCAjw04yjBhApEiwAJcvNocaaCW4IPMy4miV9pII1KoYG61jgsdIuvNPOnKhsFnXfSanKiEPT7RoC7l4QAvD_BwE#ellipsis-metavariables

brandonspark:
yeah, i think we're not very consistent about these names. ive been wondering this myself... you see `named ellipsis` in our docs here as well: https://semgrep.dev/docs/writing-rules/generic-pattern-matching/#handling-line-based-input

nmote (approved):
I didn't realize we were inconsistent with the naming. At some point we should probably clean that up. Anyway, this looks good!

PR #417, opened 2023-05-30
C#: support ellispis in enum declarations

aryx:
test plan:
test file included
./test-lang c-sharp

### Security


aryx:
This will help https://github.com/returntocorp/semgrep/issues/7914

PR #418, opened 2023-06-13
fix! make C# builds work in CI

brandonspark:
This PR updates C# again, but incorporates the changes I made in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/309 that will hopefully make builds work again.

While I'm here, I also fixed Semgrep constructs which should be usable as standalone expression statements. This is because in C#, expression statements used to just be an expression and semicolon, but they made it more restrictive, by restricting the kinds of expressions that can appear. That means things like
```
$X;
```
no longer parse. So we had to add those in.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
@@ -86,6 +86,10 @@ module.exports = grammar(standard_grammar, {
         previous,
         prec.right(100, seq($.ellipsis, ';')),  // expression ellipsis
         prec.right(100, $.ellipsis),  // statement ellipsis
+        seq($.deep_ellipsis, ';'),
+        seq($.member_access_ellipsis_expression, ';'),
+        seq($._semgrep_metavariable, ';'),
+        seq($.typed_metavariable, ';')
       );
     },
 
--- a/lang/Makefile
+++ b/lang/Makefile
@@ -16,6 +16,7 @@ SUPPORTED_TS_LANGUAGES = \
   cairo \
   clojure \
   cpp \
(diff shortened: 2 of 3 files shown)


aryx:
I can't even see the PR on tree-sitter-csharp ... github says it's too big and don't display the diff.

brandonspark:
image

no loss in parse rate

lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
```diff
@@ -86,6 +86,10 @@ module.exports = grammar(standard_grammar, {
         previous,
         prec.right(100, seq($.ellipsis, ';')),  // expression ellipsis
         prec.right(100, $.ellipsis),  // statement ellipsis
+        seq($.deep_ellipsis, ';'),
```
aryx, line 89: Why do we need that? deep_ellipsis is part of an expression no? so it should work without this extra case.

aryx (approved):

lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
brandonspark: It is, but the C# grammar has refined it so that `seq(expression, ';')` is no longer permitted, but instead an explicit `expression_statement_expression`, which outlines the specific expressions which are allowed to appear before a semicolon.

A general expression is not one of those, so deep ellipsis also becomes one of those.

aryx: Can you add a comment in the code explaining why we need this duplication?
Otherwise in 6months we will have forgotten.

PR #419, opened 2023-06-15
Solidity: add metavariables for versions

PassKeyRa:
### Description

Now it is not possible to find specific Solidity versions in the code using metavariables. However, sometimes it is necessary to define a version or range of supported Solidity versions to check for some security rules (for example, checking for overflow/underflow, for version-specific bugs - [link](https://blog.soliditylang.org/category/security-alerts/), simple outdated versions checks). 

With this update it should be possible to match the version:

```yaml
- pattern: pragma solidity >= $VER;
```

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-solidity/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-solidity/grammar.js
@@ -32,8 +32,21 @@ module.exports = grammar(base_grammar, {
           );
         },
 
-      // Metavariables. No need to patch the identifier rule because
-      // Solidity already accepts '$' as part of an identifier
+        // Metavariables. No need to patch the identifier rule because
+        // Solidity already accepts '$' as part of an identifier
+
+        // Metavariables for Solidity versions
+        _pragma_version_constraint: ($, previous) => {
+            return choice(
+                previous,
+                seq(
+                    optional(
+                        $.solidity_version_comparison_operator
+                    ),
+                    $.identifier
+                ),
+            )
+        },
       
       // Ellipsis
         _expression: ($, previous) => {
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-solidity/grammar.js
```diff
@@ -32,8 +32,21 @@ module.exports = grammar(base_grammar, {
           );
         },
 
-      // Metavariables. No need to patch the identifier rule because
-      // Solidity already accepts '$' as part of an identifier
+        // Metavariables. No need to patch the identifier rule because
+        // Solidity already accepts '$' as part of an identifier
+
+        // Metavariables for Solidity versions
+        _pragma_version_constraint: ($, previous) => {
```
aryx, line 39: Did you try instead to override solidity_version and just allow an identifier there?

aryx (approved):

lang/semgrep-grammars/src/semgrep-solidity/grammar.js
PassKeyRa: Yes, I tried `choice(previous, $.identifier)` for solidity_version, but it somehow becomes `(solidity_version (identifier))` instead of just identifier. So, not sure that it will work

aryx:
I've pushed a new version of semgrep-solidity you should be able to pull from the semgrep repo.

PassKeyRa:
@aryx thank you for the merge! I think, the semgrep-solidity submodule also needs to be updated here https://github.com/returntocorp/semgrep/tree/develop/languages/solidity/tree-sitter

aryx:
This will help https://github.com/returntocorp/semgrep/issues/8104

aryx:
update here: https://github.com/returntocorp/semgrep/pull/8105

PassKeyRa:
Thanks!

PR #420, opened 2023-06-16
fix: c# top level ellipses

brandonspark:
The C# grammar is too restrictive when it comes to top-level compilation units. It expects certain top-level constructs in a given order, and does not allow intermingling. This is too restrictive, particularly in the case of ellipses, because we expect that only global statements are ellipses.

This PR allows ellipses to occur at the top-level in a few more cases, particular around namespace declarations.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-c-sharp/grammar.js
@@ -75,6 +75,29 @@ module.exports = grammar(standard_grammar, {
       );
     },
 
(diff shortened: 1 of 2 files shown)


aryx (commented):
Why the first ellipsis is a global_stmt and not the last one?

brandonspark:
> Why the first ellipsis is a global_stmt and not the last one?

The first one falls under global statements, but then there's a namespace declaration, which also permits an ellipsis (via the type declaration change). More generally, a global statement cannot appear after a namespace declaration, in the grammar.

aryx (approved):

PR #422, opened 2023-06-26
Adding support for metavariables for python to move towards parsing patterns via tree-sitter 

SophiaSR:
As per description, this adds support for parsing metavariables as well to move towards parsing more python in tree sitter rather than menhir. Rather than creating a separate metavariable term and adding it to identifiers, this instead overrides the regex that is defined in the original tree-sitter grammar, this is needed since currently in the original tree-sitter grammar, identifier is a terminal, and thus can't do the usual choice/previous shadowing definition. 
### Security


aryx (approved):
You can put your PR comment in the code instead, next to the identifier.

PR #423, opened 2023-06-29
Update core (now deals with MISSING nodes)

mjambon:
This uses an updated runtime library

test plan:
* the CI tests check that newly-generated code works
* to check compatibility with older generated code: `cd lang && ./test-lang ocaml && ./release --dry-run ocaml` and inspect the git diffs proposed for release (result: no change in OCaml or C code). For a real test, update the ocaml-tree-sitter-core submodule in the semgrep repo and check that it builds.

### Security


aryx:
we have some regressions.

mjambon:
> we have some regressions.

I fixed the C++ problem last night. It was due to a MISSING node that's now being reported as it should. So, it's good. I made a full report here: https://github.com/returntocorp/semgrep/issues/8189

I'm looking at the rest now.

PR #426, opened 2023-07-03
feat(rust): ellipsis metavariables

brandonspark:
Everything but the last two test cases are just `tree-sitter test -u`.

### Security


ihji (approved):
LGTM 👍

PR #428, opened 2023-07-06
add typed metavar syntax extension to rust

ihji:
### Security


brandonspark (approved):
Looks good! Thanks for the test!

brandonspark:
You can run `./release rust` from the `lang` directory, after this is merged, in order to update the `semgrep-rust` submodule. Ideally we could merge https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/426 before that, however, since I currently have an outstanding change to `semgrep-rust`.

PR #430, opened 2023-07-07
feat(julia): add metavariables as idents

brandonspark:
## What:
This PR adds in metavariables as identifiers in Julia.

This is difficult because Julia allows interpolation expressions, which essentially look exactly like metavariables, albeit with lowercase identifiers allowed. We chose to just allow identifiers to be prefixed with a dollar sign, which will permit some things which are not metavariables, such as `$foo`. We can sort this out in generic translation.

We do this because only allowing regular expressions which are metavariables can cause us to fail to parse some files. For instance, doing the classic metavariable regex will cause something like `$Pack` to be parsed as two tokens, that being `$P`, an identifier, and `ack`, also an identifier. This will fail to produce a parse tree.

If we do this, but we lower the precedence on the token, it will no longer cause this issue, but it will also fail to lex metavariable tokens if both an identifier and an interpolation expression are possible -- it will prefer the interpolation expression. This is undesirable, so we just decided to allow an arbitrary identifier.

## Xfails

This PR also adds in "xfails" for `make test` and `test-lang`.

Basically, for Julia specifically, `make test` will warn that the tests may fail, and if it's the result of just replacing `interpolation_expression` with `identifier`, you should consider your tests passing. 

For `test-lang`, we keep around an array of all the xfailed languages, and if they fail, we print a message saying to inspect the output and make sure it looks OK. It also causes the tests to pass, for that language. This relies on someone actually properly testing that the output looks OK before committing.

This means we don't need to mess with `tree-sitter test` or the test files, we can just ensure that they break only in the way we expect them to. It would be nice to do something more in-depth, but I think this is the lowest friction way to ensure we can make progress.

## Test plan:
`make test` in the `semgrep-julia` repo warns about the failing test:
image

`test-lang` warns about the xfail language at the very end:
image

### Security


lang/semgrep-grammars/src/semgrep-julia/grammar.js
```diff
@@ -20,6 +20,28 @@ module.exports = grammar(base_grammar, {
...
+          // 1) this would result in cases where we tokenize as an identifier or
+          // an interpolation expression where we don't want to, depending on how
+          // we set up the precedence.
+          // 2) this causes tokenization issues with interpolation expression that
+          // have prefixes which are metavariables, such as $Pack, which is parsed
+          // as two tokens -- an identifier "$P" and an identifier "ack".
+
+          // It's better to allow anything prefixed with a dollar sign to be an identifier,
+          // then sort it out in generic translation.
+          seq("$", previous)
```
mjambon, line 41: This allows whitespace (extras) after the `$`. I would try using `token.immediate()` to prevent this.

brandonspark:
@mjambon @aryx  thoughts?

lang/semgrep-grammars/src/semgrep-julia/grammar.js
```diff
@@ -20,6 +20,28 @@ module.exports = grammar(base_grammar, {
...
+          // metavariables more faithfully, but either:
+          // 1) this would result in cases where we tokenize as an identifier or
+          // an interpolation expression where we don't want to, depending on how
+          // we set up the precedence.
+          // 2) this causes tokenization issues with interpolation expression that
+          // have prefixes which are metavariables, such as $Pack, which is parsed
+          // as two tokens -- an identifier "$P" and an identifier "ack".
+
+          // It's better to allow anything prefixed with a dollar sign to be an identifier,
+          // then sort it out in generic translation.
```
aryx, line 40: Please add a comment also here about the fact that this leads to new parsing failure and this xfail hack.
Maybe you should show an example of interpolation expression and show how now
this is parsed differently.

lang/semgrep-grammars/src/semgrep-julia/prep
```diff
@@ -1 +0,0 @@
-../prep.common
```
aryx, line 1: We can't do that. It's just not scalable. We need to find a better solution.

lang/semgrep-grammars/src/semgrep-julia/grammar.js
```diff
@@ -20,6 +20,28 @@ module.exports = grammar(base_grammar, {
...
+        choice(
+          previous,
+          // We allow an identifier to be the same as an identifier, but prefixed
+          // with a dollar sign.
+          // We could use a more specific regular expression, which would match
+          // metavariables more faithfully, but either:
+          // 1) this would result in cases where we tokenize as an identifier or
+          // an interpolation expression where we don't want to, depending on how
+          // we set up the precedence.
+          // 2) this causes tokenization issues with interpolation expression that
```
aryx, line 35: This should not be the case, we could use a regexp that ensure we stop at a word boundary (\b in Perl regexp).
I'm sure there's a way in tree-sitter to accept $P but not $Pack.

lang/semgrep-grammars/src/semgrep-julia/grammar.js
```diff
@@ -20,6 +20,28 @@ module.exports = grammar(base_grammar, {
...
     semgrep_ellipsis: $ => '...',
 
+    // Metavariables
+    identifier: ($, previous) => {
+      return token(
+        choice(
+          previous,
+          // We allow an identifier to be the same as an identifier, but prefixed
+          // with a dollar sign.
+          // We could use a more specific regular expression, which would match
```
aryx, line 30: I would vote for a better regexp with $[A-Z]+
at a word boundary.

lang/test-lang
```diff
@@ -6,6 +6,11 @@ set -eu -o pipefail
 
 prog_name=$(basename "$0")
 
+xfail_langs=('julia')
```
aryx, line 9: nah, we don't want all those hacks.
There has to be a better way.

aryx (changes requested):

bitcoinfinancier (approved):

lang/semgrep-grammars/src/semgrep-julia/grammar.js
brandonspark: @aryx Little bit of an issue:
image

https://github.com/tree-sitter/tree-sitter/blob/62e96c9f61af086641699a8848cf881d27e9afdf/cli/src/generate/prepare_grammar/expand_tokens.rs#L255

It seems that `tree-sitter` does not support word boundary assertions.

brandonspark: @aryx Some more problems:
`identifier` in the Julia grammar is what is known as the "word token", which is a special hard-coded token that is meant to disambiguate identifiers which may also be keywords (see https://tree-sitter.github.io/tree-sitter/creating-parsers#keyword-extraction)

 Notably, that token is not allowed to be a non-terminal at the rule level. It can only be a single token.

This might seem OK, because tokens themselves can be `choice`s of other token elements. The problem is that they cannot be `Symbol`s, which includes external symbols. This means I cannot make `identifier` the union of the old `identifier` and an external token, which was how I thought I might fix this. I can make `identifier` itself external, but that means I would have to re-implement it in C, which is a very complicated regex.

In summary, this is a really hard grammar to hack.

brandonspark:
@maxbrunsfeld Do you know how hard it would be to support assertions in the regex grammar preparation step? Specifically, this case: https://github.com/tree-sitter/tree-sitter/blob/62e96c9f61af086641699a8848cf881d27e9afdf/cli/src/generate/prepare_grammar/expand_tokens.rs#L255

It would make what I'm doing a lot easier.

PR #436, opened 2023-07-18
feat(promql): ellipsis for binary expressions

MichaHoffmann:
### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-promql/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-promql/grammar.js
@@ -39,6 +39,31 @@ module.exports = grammar(base_grammar, {
 
     function_args: ($) =>
       seq("(", commaSep(choice($.semgrep_ellipsis, $._query)), ")"),
+
+    binary_expression: ($) => {
+      const table = [
+        [6, choice("^")],
+        [5, choice("*", "/", "%")],
+        [4, choice("+", "-")],
+        [3, seq(choice("==", "!=", ">", ">=", "<", "<="), optional("bool"))],
+        [2, choice("and", "or", "unless")],
+        [1, choice("atan2")],
+      ];
+
+      return choice(
+        ...table.map(([precedence, operator]) =>
+          prec.left(
+            precedence,
+            seq(
+              choice($.semgrep_ellipsis, $._query),
+              operator,
+              optional($.binary_grouping),
+              choice($.semgrep_ellipsis, $._query),
+            ),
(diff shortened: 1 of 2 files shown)


MichaHoffmann:
```
Your branch is up to date with 'origin/main'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   fyi/semgrep-grammars/src/semgrep-promql/grammar.js
	modified:   fyi/tree-sitter-version
	modified:   lib/Boilerplate.ml
	modified:   lib/CST.ml
	modified:   lib/Parse.ml
	modified:   lib/parser.c

[dry run] git commit -m Release files generated by ocaml-tree-sitter 4374124.
[dry run] git push origin main

```

lang/semgrep-grammars/src/semgrep-promql/grammar.js
```diff
@@ -39,6 +39,31 @@ module.exports = grammar(base_grammar, {
 
     function_args: ($) =>
       seq("(", commaSep(choice($.semgrep_ellipsis, $._query)), ")"),
+
+    binary_expression: ($) => {
```
aryx, line 43: Did you try instead to allow '...' for _query instead?
So that we don't need special treatment for binary_expression?

MichaHoffmann: Can we go with the special treatment for now? Allowing `...` for _query in general leads to some conflicts that i cannot yet grasp the implications of.

aryx: ok. Maybe you could add a TODO that it would be better to generalize and your current findings about why we get conflicts
by doing it the clean way.

aryx: The problem I have with your current solution is that it's copy pasting code from the original grammar in table, and so if one modify the original grammar to add an operator, we will not get a warning about it.
Maybe we could export table in the original grammar so that it could simply be referenced here, instead of copy-pasted.

aryx:
cc @mjambon is the git error above is related to another error you recently had in git?

lang/semgrep-grammars/src/semgrep-promql/grammar.js
MichaHoffmann: I understand; do you mind if i clean it up as followup? ill try writing the grammar friendlier to semgrep now that i have a better understanding hwo it works.

PR #444, opened 2023-08-23
feat: update ruby with Semgrep constructs

brandonspark:
This PR updates `tree-sitter-ruby` and augments the Ruby grammar with Semgrep constructs.

One difficulty that @emjin ran into while trying to do this same thing was to differentiate range expressions `e1 ... e2` from instances of Semgrep ellipses. This is indeed difficult if you try to strictly manipulate tokens and nonterminals at the `grammar.js` level, because you need an additional level of context to differentiate these cases.

To help us differentiate these properly, we end up giving Ruby the Kotlin treatment, by maintaining a slightly augmented scanner. This is because we lex `...` as a `$.semgrep_metavariable` _only_ when the `...` is followed (modulo whitespace) by a newline, which probably shouldn't happen for an inline range expression.

For parameters and function arguments, this already exists as a "forward parameter", so we just need to take care of it during Generic translation. For dot access ellipses, we just allow identifiers to also be ellipses.

### Security


lang/semgrep-grammars/src/semgrep-ruby/src/scanner.cc
```diff
@@ -0,0 +1,1174 @@
...
+  }
+
+  bool has_leading_whitespace;
+  vector<Literal> literal_stack;
+  vector<Heredoc> open_heredocs;
+};
+
+}  // namespace
+
+extern "C" {
```
aryx, line 1148: maybe we should store the change also as a patch, so if anyone has to update scanner.cc and pull the latest, they
can just copy the new one and apply the patch.

aryx: or maybe we should just store the patch and update 'prep' to do this patching on the original file.

aryx (approved):
great work

PR #446, opened 2023-08-24
ruby fixes for menhir parsing parity

brandonspark:
Checking the `...` for being followed by a newline didn't take into account EOF. This PR fixes that (and also guards the semgrep ellipsis logic by whether it is currently a valid symbol)

### Security


aryx (approved):
Did you have time to think about my patching approach to scanner.cc instead
of full copy?

brandonspark:
@aryx i did, im open to the idea (i did something similar at Facebook for the purposes of automatically patching type stubs for library functions)

i think it would be fairly clean, ill make a ticket and try to fit it in in the near future. hopefully shouldn't be too hard.

brandonspark:
@aryx appreciate another look, I made more changes

lang/semgrep-grammars/src/semgrep-ruby/grammar.js
```diff
@@ -69,6 +69,11 @@ module.exports = grammar(standard_grammar, {
     $.semgrep_ellipsis_followed_by_newline
   ]),
 
+  conflicts: ($, previous) => previous.concat([
+    [$.keyword_pattern, $.pair]
```
aryx, line 73: would be nice each time we add a conflict to have a short comment before giving an example of such conflict.

lang/semgrep-grammars/src/semgrep-ruby/grammar.js
```diff
@@ -97,33 +103,98 @@ module.exports = grammar(standard_grammar, {
...
-        /\$[A-Z_][A-Z_0-9]*/,
-        // Need a high precedence here, so that "$...X" is not parsed
-        // as a range expression from $ to X.
-        token(prec(1000, /\$\.\.\.[A-Z_][A-Z_0-9]*/)),
-        // Same here, so it works within dot accesses.
-        alias(token(prec(1000, /\.\.\./)), $.semgrep_ellipsis)
-     ))
-    ),
+    identifier: ($, previous) =>
+      token(prec(-1,
```
aryx, line 113: why the token is now outside the choice?
Could be useful to add a comment.

lang/semgrep-grammars/src/semgrep-ruby/grammar.js
```diff
@@ -97,33 +103,98 @@ module.exports = grammar(standard_grammar, {
...
+    // Needed for "Class with ellipses"
+    constant: ($, previous) =>
+      choice(previous,
+        $.semgrep_metavariable,
+      ),
+
+    // needed for "Hash Pattern Ellipsis"
+    keyword_pattern: ($, previous) =>
+        choice(previous,
+          alias('...', $.semgrep_ellipsis)
```
aryx, line 154: why do we need aliases? Can't just use $.semgrep_ellipsis?

aryx (approved):
Nice work!

PR #447, opened 2023-08-29
add underscore back to julia metavars

brandonspark:
The previous regex for Julia metavars didn't have underscores. Let's fix that.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-julia/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-julia/grammar.js
@@ -29,6 +29,13 @@ module.exports = grammar(base_grammar, {
 
     deep_expression: $ => seq("<...", $._expression, "...>"),
 
+    // Note that this is actually slightly more permissive than a Semgrep
+    // metavariable usually is. This is fine, because having a slightly more
+    // permissive grammar is OK, we will just dispatch in the Generic
+    // translation.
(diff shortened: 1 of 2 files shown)


aryx (approved):

lang/semgrep-grammars/src/semgrep-julia/test/corpus/semgrep.txt
```diff
@@ -219,7 +220,8 @@ end
       (vector_expression))
     (do_clause
       (parameter_list
-        (identifier))
+        (identifier
+          (semgrep_extended_metavariable)))
       (return_statement
         (identifier)))))
```
aryx, line 227: maybe could have added a test with something that is not a metavariable but a valid julia identifier?

lang/semgrep-grammars/src/semgrep-julia/grammar.js
```diff
@@ -29,6 +29,13 @@ module.exports = grammar(base_grammar, {
 
     deep_expression: $ => seq("<...", $._expression, "...>"),
 
+    // Note that this is actually slightly more permissive than a Semgrep
+    // metavariable usually is. This is fine, because having a slightly more
+    // permissive grammar is OK, we will just dispatch in the Generic
```
aryx, line 34: maybe give an example here of an identifier that is not a metavar?

PR #448, opened 2023-09-07
update c and cpp

brandonspark:
This PR just updates C and C++ to the most recent.

### Security


aryx (approved):
It might be very complicated to adapt Parse_cpp_tree_sitter.ml to the many many changes that happened since we last updated tree-sitter-cpp.
We might want to do that in multiple steps; not sure you can do all at once.

PR #451, opened 2023-09-20
feat(c++): augment with semgrep constructs

brandonspark:
This PR augments C++ with Semgrep-specific constructs, as well as an alternate expression-parsing entry point. This is in the hopes that we can deprecate the old Pfff parser for Tree-sitter, which necessitates that we be able to parse patterns.

### Security


aryx (approved):
great!

PR #452, opened 2023-09-28
feat(c/c++): update to latest tree-sitter parsers

brandonspark:
This is the branch I was publishing changes from, to update the C/C++ parsers. We can now merge this, now that all the changes are through.

### Security


brandonspark:
oops, I already did this and forgot

PR #454, opened 2023-10-08
chore: bump tree-sitter-c to HEAD

amaanq:
### Security


aryx (approved):

aryx:
Thx @amaanq . I've now released a new version for semgrep-c based on your change. See https://github.com/returntocorp/semgrep-c/commit/850b018f27e54be4b7daec05101616083a2ec1cf
I've basically merged your PR in ocaml-tree-sitter-semgrep, then did cd lang; ./test-lang c; ./release c
which updated the semgrep-c repository.

aryx:
Now to take effect in semgrep itself, you need to make a PR on semgrep that use the latest version of the semgrep-c submodule (see languages/c/tree-sitter in the semgrep repo) and usually adapt Parse_c_tree_sitter.ml with the changes; in this case because it was just a change of priority, there is no needed change to Parse_c_tree_Sitter.ml. The CST remained the same so no need to adapt Parse_c_tree_sitter.ml. Still by using and linking with the latest semgrep-c/, the example phil posted should now work.

aryx:
What we usually do is we update the semgrep-c module, and then we add a test like in tests/patterns/c/misc_offset_priority.sgrep and misc_offset_priority.c showing things work better now for offset.

Let me know here or on Slack if you have more questions.

PR #459, opened 2023-11-07
chore: Update tree-sitter-rust (1/n)

nmote:
We are way behind on tree-sitter-rust. When I tried to do a big jump I ran into a bunch of problems, so I'm going to do it bit by bit.

Test plan: Automated tests

### Security


aryx:
circleCI got messed up after org rename

aryx (approved):

PR #460, opened 2023-11-21
move elixir to pro so need tree-sitter-lang-pro package and use older tree-sitter-elixir

aryx:
test plan:
./release elixir and check the semgrep-elixir now has
adjusted dune files

### Security


aryx:
see https://github.com/semgrep/semgrep/pull/9308

PR #466, opened 2024-01-12
elixir allows ellipsis in keyword/pair positions

amchiclet:
Elixir already has ellipsis, but it expects them where identifiers are expected.

But Semgrep also wants to allow ellipsis in other places. This PR allows ellipsis
where keyword/pairs are expected.

Such as function calls `foo(kwd_arg1: 0, ...)` or map items `%{item1: 0, ...}`.

Previously, these cases won't parse.

Tested with `make test`.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
@@ -19,7 +19,7 @@ module.exports = grammar(base_grammar, {
   rules: {
     // No need for _SEMGREP_EXPRESSION hack here, because Elixir allows
     // toplevel expressions.
-      
+
     // Metavariables
     identifier: ($, previous) => {
       return choice(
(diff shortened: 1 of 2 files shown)


amchiclet:
Something having to do with CircleCI's config causing the workflow to fail.

Locally ran [`the tests in CircleCI's workflow`](https://github.com/semgrep/ocaml-tree-sitter-semgrep/blob/a644def5375df9548fac8f3b4b03dacfee205e43/.circleci/config.yml#L40-L44) and they all pass.

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
```diff
@@ -33,12 +33,28 @@ module.exports = grammar(base_grammar, {
...
+    // keyword/pairs, such as
+    //   foo(some_arg: 0, ...)
+    //   %{some_item: 0, ...}
+    // Also note that now there is ambiguity whether foo(...) is an
+    // identity or pair, we set the pair rule to have a lower
+    // precedence
+    pair: ($, previous) => {
+      return prec(-1, choice(
+        previous,
+        '...',
```
brandonspark, line 49: Typically, we will define the nonterminal `$.semgrep_ellipsis` to be `"..."` and use that, rather than inlining, because it will tend to produce more descriptive typed CSTs. 

Not a big deal here though, since this is the single use case in the file, so please don't go to the trouble of revising it, rereleasing `semgrep-elixir`, and having to update your other PR!

brandonspark (approved):
Awesome!

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
aryx: Better to follow Brandon's advice though. You can look at the other semgrep-xxx/grammar.js for inspiration, but it helps to be consistent in our different grammar extensions.

PR #467, opened 2024-01-16
elixir allows metavariables as keywords

amchiclet:
Previously, keywords (as in keyword/value pairs) didn't allow metavariables, so cases like 
```elixir
  foo($KEYWORD: $VALUE)
  %{$KEYWORD: $VALUE}
```
would fail to parse.

This PR adds support for that.

Also made ellipsis its own rule, as previously suggested at https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/466.

Tested with make test and added a test.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
@@ -49,16 +49,22 @@ module.exports = grammar(base_grammar, {
     // keyword/pairs, such as
     //   foo(some_arg: 0, ...)
     //   %{some_item: 0, ...}
-    // Also note that now there is ambiguity whether foo(...) is an
-    // identity or pair, we set the pair rule to have a lower
-    // precedence
     pair: ($, previous) => {
-      return prec(-1, choice(
+      return choice(
         previous,
         $.semgrep_ellipsis,
-      ));
+      );
     },
 
+    // Note that because of the pair rule above, now there is
+    // ambiguity whether the ellipsis in foo(...) is an
+    // ellipsis for positional or keyword arguments. If ... matches
+    // with the identity rule, it will be considered a positional
+    // argument. If ... matches with semgrep_ellipsis, it will be
+    // part of parsing the pair rule, which means its a keyword
+    // argument.  In practice, this doesn't matter because either
+    // way, they will become ParamEllipsis in the Generic AST
+    // anyway.
     semgrep_ellipsis: $ => prec(-1, '...'),
 
     _expression: ($, previous) => choice(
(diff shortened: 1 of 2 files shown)


amchiclet:
The workflow `build-on-commit` failed because some CircleCI thing that we need to fix.

But I [ran the steps needed for that workflow](https://github.com/semgrep/ocaml-tree-sitter-semgrep/blob/40c10f7fd202c0a756d2f0b6f2278ae4e574c2b6/.circleci/config.yml#L37-L40C27) locally and all tests passed.

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
```diff
@@ -28,6 +28,15 @@ module.exports = grammar(base_grammar, {
...
     },
 
+    _keyword: ($, previous) => {
+	return choice(
+	    ...previous.members,
+	    $.metavariable_keyword
+      );
+    },
+
+    metavariable_keyword: $ => seq($._semgrep_metavariable, /:\s/),
```
aryx, line 38: Why the required space after : ?
Is it to match what tree-sitter-elixir does for the other keywords?

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
```diff
@@ -46,10 +55,12 @@ module.exports = grammar(base_grammar, {
     pair: ($, previous) => {
       return prec(-1, choice(
         previous,
-        '...',
+        $.semgrep_ellipsis,
       ));
     },
 
+    semgrep_ellipsis: $ => prec(-1, '...'),
```
aryx, line 62: why the prec? would be good to give the intuition on why we need this directive, like giving an example
with a conflict and how prec(-1 does the right thing and gives priority to one or the other to generate the right AST.
You can look at what we do in the other semgrep-xxx/grammar.js, we sometimes explain those prec and conflicts.

aryx (approved):

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
amchiclet: Yes, exactly.

lang/semgrep-grammars/src/semgrep-elixir/grammar.js
amchiclet: Added comments to explain why we need prec and why its correct. Thanks!

amchiclet:
The CircleCI is still failing (we haven't moved to GHA yet), so I [ran the steps needed for that workflow](https://github.com/semgrep/ocaml-tree-sitter-semgrep/blob/40c10f7fd202c0a756d2f0b6f2278ae4e574c2b6/.circleci/config.yml#L37-L40C27) locally and all tests passed.

Merging now.

PR #472, opened 2024-01-26
update cpp

brandonspark:
This PR updates `ocaml-tree-sitter-semgrep` with the most recent `tree-sitter-cpp`, which includes my fix https://github.com/tree-sitter/tree-sitter-cpp/pull/239 which should fix a [grammar error we are currently experiencing](https://linear.app/semgrep/issue/PA-3190/regression-assignments-in-c-conditions-are-not-parsed-correctly).

### Security


ihji (approved):
Thanks!

aryx:
thx a lot brandon!
Appreciate.

PR #474, opened 2024-01-31
elixir allow metavariables as atoms

amchiclet:
Elixir has atoms that are pretty much identifiers that start with a colon, e.g. `:foo`.

Metavars can match expressions, but not atoms
* `$ATOM` matches `:foo`
* `:$ATOM` doesn't parse

This PR allows metavars as atoms so `:$ATOM` matches `:foo`.
 
Also fixed some indentation in some previous code.

Test plan: added test in https://github.com/semgrep/semgrep-proprietary/pull/1274

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-elixir/grammar.js
@@ -29,13 +29,22 @@ module.exports = grammar(base_grammar, {
     },
 
     _keyword: ($, previous) => {
-	return choice(
-	    ...previous.members,
-	    $.metavariable_keyword
+      return choice(
+        ...previous.members,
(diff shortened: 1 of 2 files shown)


aryx (approved):
We have other languages with atoms.
Do we support this also for Ruby?

amchiclet:
I will check and report back.

amchiclet:
Yes, this is supported this same behavior in Ruby.

amchiclet:
Ran `make setup`, `make install`, `make -C core test`, and `make lang` locally. All of them passed, so I'm merging now.

Filed a linear ticket to fix CI.

amchiclet:
cc @p4p3r here's the metavar atom parsing fix.

PR #476, opened 2024-03-26
fix(kotlin): classes with newlines

brandonspark:
## What:
This PR makes it so we can parse Kotlin classes that have a newline between the class identifier and constructor.

## Why:
Parse rate.

## How:
The problem is in cases like:
```
class Foo
constructor Bar() { ... }
```

Here, we insert an automatic semicolon between `Foo` and `constructor`. This leaves us able to parse `class Foo` as a `class_declaration`, but `constructor Bar ...` is not allowed on its own.

We simply allow `constructor Bar ...` to be a standalone statement, and resolve to stitch them together at parsing time.

### Testing

Tested with `make test` and `test-lang kotlin`

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
@@ -90,6 +90,11 @@ module.exports = grammar(standard_grammar, {
   // The former has `secondary_constructor`, which already looks like what we want to
   // add, and the second seems to be unused.
   // So we just need to fix _statement.
+
+  // A more proper fix would likely require changes to the external scanner to properly
+  // handle automatic semicolon insertion, which is the cause of this whole issue.
+  // We filed an issue to tree-sitter-kotlin to track this:
+  // https://github.com/fwcd/tree-sitter-kotlin/issues/75
   _statement: ($, previous) => choice(
     ...previous.members,
     $.partial_class_declaration,
--- a/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
@@ -91,18 +91,18 @@ module.exports = grammar(standard_grammar, {
   // add, and the second seems to be unused.
   // So we just need to fix _statement.
   _statement: ($, previous) => choice(
-    previous,
-      $.partial_class_declaration,
+    ...previous.members,
+    $.partial_class_declaration,
   ),
 
-        partial_class_declaration: $ =>     prec.left(seq(
-      optional($.type_parameters),
-      seq(optional($.modifiers), "constructor"),
-      $._class_parameters,
-      optional(seq(":", $._delegation_specifiers)),
-      optional($.type_constraints),
-      optional($.class_body)
-        )),
+  partial_class_declaration: $ => prec.left(seq(
+    optional($.type_parameters),
+    seq(optional($.modifiers), "constructor"),
+    $._class_parameters,
+    optional(seq(":", $._delegation_specifiers)),
+    optional($.type_constraints),
+    optional($.class_body)
+  )),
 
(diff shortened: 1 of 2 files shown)


lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
```diff
@@ -74,13 +74,41 @@ module.exports = grammar(standard_grammar, {
...
+  _statement: ($, previous) => choice(
+    previous,
+    prec.left(seq(
+      optional($.type_parameters),
+      seq(optional($.modifiers), "constructor"),
+      $._class_parameters,
+      optional(seq(":", $._delegation_specifiers)),
+      optional($.type_constraints),
+      optional($.class_body)
+    ))
```
amchiclet, line 102: Very nice workaround. Thanks!

To simplify stitching during CST -> generic AST (which I'm currently working on using these changes), I think grouping this into its own rule such as `partial_class_declaration` would help.

amchiclet: Wasn't sure if I should modify your PR directly or not. I created another PR that builds on top of this PR to refactor the rules a little bit because it makes it easier for the CST -> Generic AST process.

https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/477

amchiclet: Sounds good, will do, thanks!

amchiclet: Merged my commits into this branch.

amchiclet:
FYI `make lang` in Circle CI runs out of memory when parsing C sharp. Ran it locally, and it passed.

lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
```diff
@@ -74,13 +74,43 @@ module.exports = grammar(standard_grammar, {
...
+  // it such that we get interrupted in the middle of the class_declaration.
+  // To make it so we can continue, we allow everything after the class identifier
+  // to be a standalone statement in its own right. This way, we can parse both parts
+  // individually, and stitch them together at parsing time.
+
+  // We only need to amend statements here, because the consumers of _declaration are
+  // only class_member_declaration, top_level_object and _statement.
+  // The former has `secondary_constructor`, which already looks like what we want to
+  // add, and the second seems to be unused.
+  // So we just need to fix _statement.
```
nmote, line 92: Good comment, but could you also include a note about what the non-hacky way to do this would be, perhaps with a link to an issue on the tree-sitter-kotlin repo? I assume the scanner would have to be updated to change the ASI behavior?

nmote (approved):
Not a fan of all the formatting changes in the tests. I guess the "hide whitespace" setting only has to do with whitespace changes on a single line, so to review, one still has to compare these manually.

I spot-checked some and they looked the same, but I'm assuming with this review that the only behavior change is for the added test cases.

lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
amchiclet: Done.

amchiclet:
Also spot checked and arrived at the same conclusion. The only behavior change is the new test cases.

Thanks for the quick review!

brandonspark:
> Not a fan of all the formatting changes in the tests. I guess the "hide whitespace" setting only has to do with whitespace changes on a single line, so to review, one still has to compare these manually.
> 
> I spot-checked some and they looked the same, but I'm assuming with this review that the only behavior change is for the added test cases.

the formatting changes are automatically generated from `tree-sitter test -u`. it's a pain, but this slight annoyance in review now will make future reviews easier.

nmote:
In the future, I'd prefer reformatting to be split into a separate PR

PR #477, opened 2024-03-27
fix(kotlin): classes with newlines

amchiclet:
This PR refactors Brandon's PR, so the CST to Generic AST process becomes a bit easier. The rest of the description is copied and pasted from https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/476

## What:
This PR makes it so we can parse Kotlin classes that have a newline between the class identifier and constructor.

## Why:
Parse rate.

## How:
The problem is in cases like:
```
class Foo
constructor Bar() { ... }
```

Here, we insert an automatic semicolon between `Foo` and `constructor`. This leaves us able to parse `class Foo` as a `class_declaration`, but `constructor Bar ...` is not allowed on its own.

We simply allow `constructor Bar ...` to be a standalone statement, and resolve to stitch them together at parsing time.

### Security

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-kotlin/grammar.js
@@ -91,18 +91,18 @@ module.exports = grammar(standard_grammar, {
   // add, and the second seems to be unused.
   // So we just need to fix _statement.
   _statement: ($, previous) => choice(
-    previous,
-      $.partial_class_declaration,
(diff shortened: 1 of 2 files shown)


amchiclet:
I will merge commits from this PR into Brandon's original PR: https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/476#discussion_r1541492626.

Not closing this PR yet. I'd like the branch to still be alive while merging changes.

PR #478, opened 2024-03-27
feat: add codeql language

brandonspark:
This PR adds the CodeQL language to Semgrep.

### Security

- [ ] Change has no security implications (otherwise, ping the security team)

PR #479, opened 2024-04-01
feat: few more codeql things

brandonspark:
i merged the last PR #478  before pushing all my local changes

### Security


aryx (approved):

aryx:
Not sure what is the CI failure but otherwise looks good.

brandonspark:
The CI failure seems to be a consistent flaw in our CI now, since we have downgraded our runners. Building the C# parser will break our CI. See also: https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/478

PR #480, opened 2024-04-24
feat: augment C with semgrep constructs

brandonspark:
This PR augments the C grammar with Semgrep constructs, so that we can move to deprecate the `pfff` parser for C.

I went through all the `sgrep-ext`s in `Parser_cpp.mly` to make sure that we maintain parity.

Added some cpp stuff too.

### Security


aryx (approved):
I think medium-term we want to get rid of semgrep-c and just use semgrep-cpp to also parse C code.

nmote:
Seems abandoned

PR #481, opened 2024-05-07
Update tree-sitter-dockerfile

mjambon:
Uses the latest tree-sitter-dockerfile from the official repo, main branch.

test plan:
```
cd lang
./test-lang dockerfile
```
I pushed the generated code successfully to https://github.com/semgrep/semgrep-dockerfile with `./release dockerfile`.

### Security


aryx (approved):
Sure. CI is failing though.

PR #484, opened 2024-06-11
Test as many languages as possible before failing in CI

mjambon:
We have this old issue that `tree-sitter generate` requires a very large amount of memory (> 16 GB) and time (> 30 min?)
 for some grammars after their conversion by ocaml-tree-sitter. Now, the CI job still fails but it tests all the languages it can and reports the failed languages at the end.

I put `hack` back into the list of languages to be tested. The languages that fail are printed as follows:
```
*** Failed to build or test the following languages: c-sharp hack
```

### Security

### DIFF
--- a/lang/Makefile
+++ b/lang/Makefile
@@ -148,7 +148,7 @@ test:
 	set -e; \
         failed=""; \
 	for lang in $(SUPPORTED_TS_LANGUAGES); do \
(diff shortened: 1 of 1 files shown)


mjambon:
```
/bin/sh: 4: failed+= ang: not found
```
oops

brandonspark (approved):

aryx (approved):

PR #487, opened 2024-06-19
Upgrade ocaml-tree-sitter-core, upgrading tree-sitter to 0.22.6

mjambon:
We're now using tree-sitter 0.22.6 by default. This doesn't work for many of the existing parsers because they need to be upgraded. They will be listed as failed by the CI check.

See also https://github.com/semgrep/ocaml-tree-sitter-core/pull/77 and https://github.com/semgrep/semgrep-proprietary/pull/1691

### Security


mjambon:
Hmm, wrong branch. There's already a PR for upgrading tree-sitter: https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/485

PR #488, opened 2024-06-25
chore(js/ts): allow ellipses to occur in jsx elements

brandonspark:
As title

This PR also fixes the Makefile to also include the name of the dialect that it is dealing with. Previously it only printed the PWD.

### Security


aryx (approved):
LGTM.
Actually we currently have some implicit "..." in JSX attributes; even if you don't
use a ..., we match like there is some "...". I always wanted to force people
to add "...", to be more consistent, but never took the time because of fear of breaking existing rules.

Nore sure what's going on with CI though.
maybe @mjambon know.

brandonspark:
@aryx i believe it's spurious, see the job:
image

PR #490, opened 2024-07-02
Upgrade C# grammar

mjambon:
This uses the latest version of tree-sitter-c-sharp and updates the grammar extension for semgrep. `tree-sitter generate` still takes more than 16GB of RAM :-(

### Security

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ version: 2.1
 
 jobs:
 
-  build:
+  test-language:
     docker:
       - image: returntocorp/ocaml:ubuntu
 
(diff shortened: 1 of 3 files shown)


brandonspark (approved):

DrewDennison (approved):

aryx (changes requested):
I get conflicts and test regressions with this PR (I'm able to run tree-sitter on my 32GB machine).

aryx:
probably better to restart from scratch since I've made an intermediate PR that update also to a more recent tree-sitter-csharp (but not the latest one).

PR #492, opened 2024-07-02
Update to a more recent tree-sitter-c-sharp before the big refactor

aryx:
The big one is https://github.com/tree-sitter/tree-sitter-c-sharp/pull/333

test plan:
cd lang
./test-lang c-sharp

### Security


IagoAbal (approved):
stamping but tests are failing

PR #494, opened 2024-07-09
Circom

Frodan:
This PR adds support for the circom language.

Tree-sitter parser: https://github.com/Decurity/tree-sitter-circom/
### Security

### DIFF
--- a/lang/circom/fyi.list
+++ b/lang/circom/fyi.list
@@ -1,0 +1,3 @@
+semgrep-grammars/src/tree-sitter-circom/LICENSE
+semgrep-grammars/src/tree-sitter-circom/grammar.js
+semgrep-grammars/src/semgrep-circom/grammar.js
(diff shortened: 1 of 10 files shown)


aryx (approved):
Would be good to have a non-empty corpus/semgrep.txt test file, but you can add that in another PR.

lang/circom/fyi.list
```diff
@@ -0,0 +1,3 @@
+semgrep-grammars/src/tree-sitter-circom/LICENSE
```
aryx, line 1: This LICENSE file is missing in your tree-sitter-circom repository, which prevents me to run
the ./release circom script to populate the https://github.com/semgrep/semgrep-circom
Please add this file, update the submodule, and make a PR, so I can then run ./release circom

Frodan: Thank you, fixed in PR: https://github.com/semgrep/ocaml-tree-sitter-semgrep/pull/495

PR #501, opened 2024-07-29
Fix Move on sui grammar.js

aryx:
test plan:
cd lang
./test-lang move-on-sui
now works, without linking error

### Security


chargarlic (approved):

akuhlens (approved):
🙈

PR #503, opened 2024-08-06
Feat(move-on-aptos): v2 features

ArArgon:
- Move on aptos v2 features:
  - Enum
  - Match pattern
  - Positional fields
  - Package visibility
  - Alternative syntax for struct and enum
- Bug fixes
  - `vector[...]` is correctly parsed
  - Change `return`'s precedence
  - Some grammar clean-ups

### Security


nmote (approved):
I'm not familiar with move on aptos, but this looks reasonable and well-tested to me.

nmote:
I've updated https://github.com/semgrep/semgrep-move-on-aptos with these changes.

PR #504, opened 2024-08-27
feat(go): basic semgrep support

brandonspark:
Added ellipses, metavars, deep metavars, and typed metavars.

### Security


aryx:
We didn't support those features before??
How people managed to write semgrep rules for Go?
We were handling that in the menhir parser?

brandonspark:
> We didn't support those features before?? How people managed to write semgrep rules for Go? We were handling that in the menhir parser?

well, i thought we didn't, but we do. in any event, i think it would be nice if we standardized on tree-sitter

aryx:
Why did you make this PR? What was failing that pushed you add those features in the tree-sitter parser?
I'm all for it, I agree we should standardize on tree-sitter for parsing patterns, but I was just curious what
led you to this PR?

aryx (approved):

brandonspark:
> Why did you make this PR? What was failing that pushed you add those features in the tree-sitter parser? I'm all for it, I agree we should standardize on tree-sitter for parsing patterns, but I was just curious what led you to this PR?

the actual context is that I ended up getting a parse error in the Playground for a particular top-level ellipsis example, and figured that our Menhir parser must have just missed that case. i figured that it wouldn't take long to add full `tree-sitter` support for Go so i decided to just do that.

it turns out the exception i was getting was erroneous. it's because i was using Turbo Mode, and for unknown reasons, sometimes Turbo Mode throws parsing exceptions on well-parsing code

brandonspark:
you can actually reproduce here:
https://semgrep.dev/playground/s/106QG

brandonspark:
(this might not even be turbo mode, it could be structure mode)

PR #505, opened 2024-09-05
fix(go): bump tree-sitter parser for go to resolve bug

yosefAlsuhaibani:
### Security


lang/semgrep-grammars/src/semgrep-go/grammar.js
```diff
@@ -10,18 +10,28 @@ module.exports = grammar(base_grammar, {
...
+    semgrep_ellipsis: $ => "...",
     semgrep_deep_ellipsis: $ => seq("<...", $._expression, "...>"),
 
+    // 
+    _statement: ($, previous) => choice(
+      previous,
+      prec(1,$.semgrep_ellipsis_metavar),
+      prec(1,$.semgrep_deep_ellipsis),
+      prec(1,$.semgrep_ellipsis)
+    ),
```
yosefAlsuhaibani, line 33: I am not sure if there's a better way to go about doing this via tree-sitter, but a conflict that is awkward to resolve is 
```go
if (true) {
  ...
};
```
Won't be parsed as `(if (...) (block (semgrep_elipsis)))` but instead `(if (...) (block (expression statement (semgrep_elipsis))))`; this is what this is trying to resolve

lang/semgrep-grammars/src/semgrep-go/test/corpus/semgrep.txt
```diff
@@ -123,11 +123,12 @@ foo(..., 5)
...
-    (identifier)
-    (argument_list
-      (semgrep_ellipsis)
-      (int_literal))))
+  (expression_statement
+    (call_expression
+      (identifier)
+      (argument_list
+        (semgrep_ellipsis)
+        (int_literal)))))
```
yosefAlsuhaibani, line 131: However this should be expected, as the function call `foo(...,5)` is an expression in a statement position

brandonspark (approved):
it is a hack, but i've done worse things

PR #511, opened 2024-09-18
Circom grammar update

Frodan:
Updating tree-sitter and semgrep grammar for circom language.


aryx:
@mjambon is in the middle of a big refactoring in ocaml-tree-sitter-semgrep but once his PR is merged we can probably merge this one.

aryx:
@mjambon can we merge this? Are you done with your tree-sitter version refactoring?

Frodan:
Hi, any updates on merging this?

aryx (changes requested):
Sorry for the late review, we had an offsite last week.
Can you add some test corpus for those new supported constructs?
Also was it possible instead to modify  the original grammar so that
we don't have to copy so much from main_component_definition and component_declaration?
like if you introduce an intermediate rule for what is after the =, like main_component_value, then it's possible to override just that.

nmote:
This looks to be abandoned. Feel free to open a new PR (or reopen this one if that's possible) if you'd still like to contribute.

PR #512, opened 2024-09-19
Add another git submodule for C# called c-sharp-pro

mjambon:
This uses `tree_sitter_c_sharp_pro_` prefix in the C functions to avoid conflicts with the other C# parser so they can be linked into the same executable.

### Checklist


lang/c-sharp-pro/ConvexHull.cs
```diff
@@ -0,0 +1,322 @@
+/*
```
nmote, line 1: What's this source file doing here? Should it be in a test directory?

lang/language-variants-0.20.6
```diff
@@ -1,6 +1,5 @@
 apex
 bash
-c-sharp
```
nmote, line 3: Is this related to adding the new submodule?

lang/semgrep-grammars/src/semgrep-c-sharp-pro/grammar.js
```diff
@@ -0,0 +1,144 @@
+/*
+ * semgrep-csharp
+ *
+ * Extend the original tree-sitter C# grammar with semgrep-specific constructs
+ * used to represent semgrep patterns.
+ *
+ * The language is renamed from c-sharp to csharp to match language name
```
nmote, line 7: Looks like it's actually named `c_sharp_pro` -- is this a copy/paste error?

nmote (approved):

lang/c-sharp-pro/ConvexHull.cs
mjambon: You're right. I copied it over from the original c-sharp folder without noticing it shouldn't be here.

lang/language-variants-0.20.6
mjambon: No. I made a mistake last week when building this list. It's because the c-sharp was failing to build in CI but it was due to a lack of memory, not an incompatibility with the latest tree-sitter.

lang/semgrep-grammars/src/semgrep-c-sharp-pro/grammar.js
mjambon: yes

lang/c-sharp-pro/ConvexHull.cs
mjambon: um, it was in the correct place originally. I just did a wrong copy-paste.

mjambon: both c-sharp and c-sharp-pro now contain the same files:
```
~/ocaml-tree-sitter-semgrep/lang/c-sharp-pro $ git ls-files
Makefile
extensions.txt
fyi.list
projects.txt
test/ok/ConvexHull.cs
```

lang/semgrep-grammars/src/semgrep-c-sharp-pro/grammar.js
```diff
@@ -0,0 +1,145 @@
+/*
+ * semgrep-c-sharp-pro
+ *
+ * Extend the original tree-sitter C# grammar with semgrep-specific constructs
+ * used to represent semgrep patterns.
+ *
+ * c-sharp-pro is the same as c-sharp but tracks a more recent version
```
aryx, line 7: it's a bit annoying that we have to duplicate grammar.js ... but maybe in the longer term it's better.

aryx:
I thought it would have been simpler to use just one tree-sitter-chsarp, and update it here to the latest but update to a more
recent semgrep-c-sharp in pro, but maybe you're right it's simpler to separate them also here.

aryx:
especially if you managed to avoid naming conflicts in C.

aryx:
cc @yosefAlsuhaibani who might have to do something similar for another language (I don't remember which one).

PR #514, opened 2024-10-10
Fix C# regression in preprocessor directive parsing

mjambon:
It was due to a left-over syntax extension workaround that no longer applies and was causing conflicts.

This also fixes incorrect parsing of typed metavariables where the type is parametrized. The problem occurred just now in Semgrep and didn't have a test for this specific case (e.g. `(List<T> $X)` was parsed as `(List < T) > $X)`).

Integration work in progress at https://github.com/semgrep/semgrep-proprietary/pull/2390

### Checklist

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-c-sharp-pro/test/corpus/semgrep.txt
+++ b/lang/semgrep-grammars/src/semgrep-c-sharp-pro/test/corpus/semgrep.txt
@@ -260,6 +260,52 @@ __SEMGREP_EXPRESSION
         (identifier))
       (argument_list))))
 
+===============================================================================
+Typed metavariable 2
+===============================================================================
+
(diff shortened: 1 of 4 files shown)


lang/semgrep-grammars/src/semgrep-c-sharp-pro/test/corpus/semgrep.txt
```diff
@@ -376,3 +376,62 @@ namespace $N;
...
+        (preproc_if
+          (identifier)
+          (expression_statement
+            (invocation_expression
+              (identifier)
+              (argument_list)))
+          (expression_statement
+            (invocation_expression
+              (identifier)
+              (argument_list))))))))
```
mjambon, line 437: ^ This test was failing

aryx (approved):

PR #524, opened 2025-03-05
Adding tree-sitter-toml

aryx:
This can be useful to parse certain lockfiles (e.g. uv.lock)

test plan:
cd lang; ./test-lang toml

### Checklist

### DIFF
--- a/lang/semgrep-grammars/src/semgrep-toml/grammar.js
+++ b/lang/semgrep-grammars/src/semgrep-toml/grammar.js
@@ -1,0 +1,29 @@
+/*
+  semgrep-toml
+
+  Extends the standard toml grammar with semgrep pattern constructs.
+*/
+
+const base_grammar = require('tree-sitter-toml/grammar');
+
+module.exports = grammar(base_grammar, {
+  name: 'toml',
+
+  conflicts: ($, previous) => previous.concat([
+  ]),
+
+  /*
+     Support for semgrep ellipsis ('...') and metavariables ('$FOO'),
+     if they're not already part of the base grammar.
+  */
+  rules: {
+  /*
+    semgrep_ellipsis: $ => '...',
+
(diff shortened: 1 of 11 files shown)


lang/semgrep-grammars/src/semgrep-toml/grammar.js
```diff
@@ -0,0 +1,29 @@
...
+  rules: {
+  /*
+    semgrep_ellipsis: $ => '...',
+
+    _expression: ($, previous) => choice(
+      $.semgrep_ellipsis,
+      ...previous.members
+    ),
+  */
+  }
```
bkettle, line 28: this won't currently support writing rules for toml, right? Obviously we don't need that for lockfile parsing, but wonder if we should make it obvious in a comment in semgrep-core when we consume this or something

bkettle (approved):

nmote:
Cool! I like the idea of consolidating lockfile parsing on tree sitter. It's probably a bit clunky to consume the generic AST for that purpose, but I bet it's not too bad and it's definitely less work than maintaining a whole separate parser.

kopecs:
@nmote We can just use the CST, no?

bkettle:
> @nmote We can just use the CST, no?

I don't know if it's the best plan, but when I was playing with this earlier the CST was pretty gross so I was attempting to translate to a romp-specific intermediate AST and just use that in our lockfile parsers (without ever translating to the generic AST)

nmote:
Sure, but you are still doing most of the work (both up front and maintenance) of just integrating it into Semgrep normally at that point. I don't think the tree sitter CST is going to be more pleasant to consume than the generic AST for these purposes anyway.

aryx:
(so what Ben started a few months ago).

nmote (approved):
Just realized the code owner reviews are required (not just advised) in this repo.

PR #526, opened 2025-03-10
feat: add `requirements.txt` lang

brandonspark:
As title, adds the `tree-siter` parser for `requirements.txt`.


aryx:
would be good to add it to the Makefike so at least some of its tests are run in CI.

aryx (approved):

aryx:
@salolivares just FYI. Most of the files in this PR were actually auto-generated by a script, in lang/add-simple-lang.
Then brandon used another script lang/release
to automatically generate the OCaml binding code and put it in the semgrep-requirements repo here: https://github.com/semgrep/semgrep-requirements
and this new repo can then be used as a submodule in the semgrep-pro repo.

PR #528, opened 2025-03-13
Typescript: allow ellipsis in class body

ihji:
### Checklist


lang/semgrep-grammars/src/semgrep-typescript/common/semgrep-ext.js
```diff
@@ -39,6 +39,34 @@ module.exports = {
       $.optional_parameter,
     ),
 
+    class_body: $ => seq(
+      '{',
+      repeat(choice(
```
aryx, line 44: would have been better to have an intermediate `class_element` rule in the original grammar so
we would not have to repeat so much from the original grammar.

aryx: An issue is that if tomorrow the original grammar change class_body and add some constructs in it,
we have no way to statically check that we should also update our class_body :(
if instead we had an intermediate class_element, then could just extend the previous class_element
and use a `$.previous` like we do for the other rules.

lang/semgrep-grammars/src/semgrep-typescript/common/semgrep-ext.js
```diff
@@ -39,6 +39,34 @@ module.exports = {
       $.optional_parameter,
     ),
 
+    class_body: $ => seq(
```
aryx, line 42: At least add a big coupling comment here saying that each
time we update the grammar we should make sure we update this class_body to reflect changes
in the original grammar.

aryx (changes requested):

lang/semgrep-grammars/src/semgrep-typescript/common/semgrep-ext.js
ihji: Make sense. I've added the comment.

aryx (approved):
Would be great if you make a PR to the tree-sitter-typescript with this class_element introduction. So at least at the next update maybe it will be there and we can simplify this.

PR #542, opened 2025-06-12
feat: add json language

brandonspark:
Our old JSON parser uses the Menhir JS/TS parser, which we are trying to deprecate. However, the new `tree-sitter` parser cannot directly parse a JSON, since it is expecting a program. We should just make a new JSON parser, which is likely to also be faster.


yosefAlsuhaibani (approved):
Yes

brandonspark:
oh this is approved already mb, ignore @dijkstracula

PR #571, opened 2026-03-31
feat: add branch flag to release script

kopecs:
Adds a new `--branch $BRANCH` flag to the release script at `lang/release` so
that we do not have to rely on bypassing branch protection, which we've added
to many repos' main branch when performing a release. This also makes it easier
for external contributors to open PRs to those repos directly (if they exist),
which is a point of friction occasionally (if we want to suggest this, we would
also need to update the main contributing docs).

### Checklist

### DIFF
--- a/lang/release
+++ b/lang/release
@@ -20,7 +20,7 @@ for 'typescript' and for 'tsx'. You need to release both:
   ./$prog_name typescript
   ./$prog_name tsx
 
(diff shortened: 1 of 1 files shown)


dijkstracula (approved):

lang/release
kopecs: @dijkstracula here; we update the variable above.

PR #586, opened 2026-04-30
[ruby] fix Semgrep pattern parsing for instance/class variable metavars and array patterns

brandonspark:
## Summary

Augments the Semgrep Ruby grammar so two more pattern shapes parse correctly:

- **LANG-485**: instance / class variable metavars `@$X` and `@@$X`
  - Upstream tree-sitter-ruby defines `instance_variable` and `class_variable` as plain tokens (not via the external scanner), so we fuse a metavar regex into each via `token(prec(1000, /@\$[A-Z_][A-Z_0-9]*/))` (and likewise for `@@`).
- **LANG-491**: array / find / hash pattern ellipsis e.g. `case x in [$A, $B, ...]`
  - Done by extending `_pattern_expr_basic` with `$.semgrep_ellipsis`. Adds a few `conflicts` entries between `_pattern_expr_basic`, `keyword_pattern`, `_primary`, and `pair` because all four can now reduce to a bare `...` in some contexts; any of those reductions is fine for the Generic translation.

**Deferred**: LANG-481 (`?` / `!` predicate-bang suffix on metavars). That requires a scanner-level / entry-point lexer change rather than an additive grammar rule, so it is being tracked separately.

## Test plan

- [x] Added 5 new corpus tests in `test/corpus/semgrep.txt`:
  - `Instance variable metavariable` (`@$X`)
  - `Class variable metavariable` (`@@$X`)
  - `Instance and class variable metavariable assignments`
  - `Array pattern with ellipsis at end` (`1 in [$A, $B, ...]`)
  - `Array pattern with ellipsis in middle` (`1 in [$A, ..., $B]`)
- [x] All 47 semgrep corpus tests pass via `tree-sitter test`.
- [x] The two pre-existing inherited test failures (`global variables`, `alias`) are unchanged on this branch — they fail identically on `main` and are unrelated to this PR.
- [x] Spot-checked realistic patterns: classes with `@$X` / `@@$Y` ivars, `case` / `in` matches with `[$A, $B, ...]` and `{key: $V, ...}`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brandonspark:
Companion semgrep-ruby PR: https://github.com/semgrep/semgrep-ruby/pull/1
Release SHA: `3cb8330f17b62f2a0576bc1212a11ac20cbea706`

PR #598, opened 2026-06-01
chore: update gomod parser

bkettle:
There are a couple of improvements in the gomod parser, namely
https://github.com/camdencheek/tree-sitter-go-mod/commit/3fccb98fc203617bf7e9c486c16842efe031e54c#diff-9e4a9e7627751e36f15ab0e8044352fe58b6e09d70b8b4d5fae4951ab5146030,
that we want. 

### Checklist

I think this is true, but not 100%. I don't think we need any changes to the parsing code in Semgrep


brandonspark (approved):
my claude thinks this is ok

brandonspark:
> I think this is true, but not 100%. I don't think we need any changes to the parsing code in Semgrep

you're going to need to bump the submodule in `semgrep-proprietary`, not sure if you're just not considering that a "code change" per se

bkettle:
> > I think this is true, but not 100%. I don't think we need any changes to the parsing code in Semgrep
> 
> 
> 
> you're going to need to bump the submodule in `semgrep-proprietary`, not sure if you're just not considering that a "code change" per se

That happens after this is merged right? My understanding is I merge this, then run ./release to update Semgrep-gomod, then bump that submodule in Semgrep pro? Or is that backwards? The docs I could find about this are external facing and seem old

bkettle:
@brandonspark does the order of the semgrep-pro bump and this bump matter? I think not? The semgrep-pro PR is here: https://github.com/semgrep/semgrep-proprietary/pull/6452

brandonspark:
the merging of this PR actually doesn't matter whatsoever except for bookkeeping purposes. technically the only part of this that is relevant is the `./release` step, to give `semgrep-proprietary` somewhere to point its `semgrep-gomod`. if you merge or don't merge this PR, it doesn't matter to the substance of `semgrep-proprietary`'s PR

brandonspark:
but in terms of our procedure, we'd prefer this is merged _after_ merging `semgrep-proprietary`

brandonspark:
> That happens after this is merged right? My understanding is I merge this, then run ./release to update Semgrep-gomod, then bump that submodule in Semgrep pro? Or is that backwards? The docs I could find about this are external facing and seem old

oh didn't see this. yeah, you merge this _after_ running `./release` and bumping semgrep pro

bkettle:
OK, I've merged the result into semgrep-pro

PR #600, opened 2026-06-03
fix(dart): parse polyglot semgrep patterns

AlexLaroche:
## Summary

Three wrapper-grammar extensions so the polyglot semgrep patterns
parse cleanly for Dart, removing the need for Dart-specific exception
lists in `semgrep-core` (see semgrep/semgrep PR for the downstream
exception drops).

| Commit | Pattern shape | Without these fixes |
|---|---|---|
| `3b01075` | `class $X { ... }` and top-level `$V = get(); ...; eval($V);` | `_class_member_definition` rejects `...`; Dart's grammar greedily parses bare top-level assignments as function signatures and errors at `=` |
| `e8c4f61` | `$X = $O.foo(). ... .bar(). ...` | `selector` accepts only `_assignable_selector` / `argument_part` / `type_arguments` / `!` — no rule for a bare `...` between chain segments |
| `dde5fe3` | Multi-statement polyglots like `dots_stmts`, `metavar_equality_var` | `_semgrep_pattern` only covers single-expression / single-statement; multi-statement patterns fall back to whole-program parsing and Dart greedily resolves them as malformed function declarations |

Each commit adds:
- An extension rule to `lang/semgrep-grammars/src/semgrep-dart/grammar.js`
  (`semgrep_dot_ellipsis_selector`, `semgrep_statement_list`, the
  `_class_member_definition` / `_top_level_definition` arms).
- A corpus fixture under `lang/dart/test/ok/` pinning the new parse path.

## Test plan

- [x] `make` builds the regenerated parser cleanly.
- [x] All four new `lang/dart/test/ok/` fixtures parse with no errors.
- [x] Existing `lang/dart/test/ok` corpus continues to parse (no regressions on real Dart code introduced by the new arms).
- [x] Downstream consumer (`semgrep-dart` regen → `semgrep` mapper) has the four corresponding mapper arms wired up and all Dart pattern/maturity tests pass on the semgrep side (`./test -s 'dart'` → 63/63 pass).

## Notes for reviewers

- The `_top_level_definition` / `expression_statement` ambiguity introduced by accepting bare top-level assignments is resolved with an explicit `conflicts` entry; tree-sitter picks the expression-statement arm in pattern mode because that's the only one that reduces.
- `semgrep_statement_list` requires `repeat1` (at least two statements) so the single-statement case keeps using `$._statement` and avoids an ambiguity between the two arms.
- The `DotAccessEllipsis` Generic AST node already exists (used by C#'s `member_access_ellipsis_expression`); the Dart mapper reuses it verbatim.


malaverdiere:
Sorry to keep you waiting. We did a substantial repo consolidation. The [documentation](https://docs.semgrep.dev/contributing/updating-a-grammar) is also live on our website. You can create a PR and tag me directly.

PR #592, opened 2026-05-13
trivial: Added Grammar update script and Conductor configuration

malaverdiere:
Added a script to automate updating language grammars

Also added a configuration for Conductor users.

### Checklist

- [x] Any new parsing code was already published, integrated, and merged into Semgrep. DO NOT MERGE THIS PR BEFORE THE SEMGREP INTEGRATION WORK WAS COMPLETED.
- [x] Change has no security implications (otherwise, ping the security team)

PR #610, opened 2026-06-24
Grammar-update automation: propose PRs here, release to semgrep-<lang>

brandonspark:
Workflow that proposes upstream tree-sitter grammar updates, wrapping `scripts/update-grammar` (#592).

Per language: resolve latest stable tag → bump + regen corpus snapshots + `test-lang`, once, on `grammar-update/<lang>/<tag>` → if green, PR in this repo → integrate job releases that exact branch (tag/branch read from the result artifact) to `semgrep-<lang>` via PR.

`workflow_dispatch` only for now (cron scaffolded, commented out). Inputs:
- `language` — one name, or `all`/blank (pre-filtered to languages behind their latest tag)
- `run_review_agent` (default off) — on test-lang failure, run a Cursor agent driving the `fix-semgrep-grammar` skill (#619), then re-run test-lang. Off = Cursor CLI never installed, key never exported.
- `dry_run` — full pipeline, push nothing.

semgrep-proprietary integration is out of scope — stays with the existing Cursor workflow on that repo.

Tested:
- 36 unit tests (`make test-python`): tag resolution/filtering, propose failure + agent paths, result-artifact round-trip, env validation
- [live dry run of the current tip](https://github.com/semgrep/ocaml-tree-sitter-semgrep/actions/runs/29123594847) (`language=php, dry_run=true`, all 4 jobs green): php was genuinely behind, so the bump ran for real — `7d07b41` → `5b5627fa` (v0.24.2), snapshot regen, test-lang green (parse-examples pass, no Blank nodes, no corpus churn). Propose job took 10 min (core build cache hit). `result-php.json` round-tripped through the Result dataclass; the integrate job read it and gated itself off. Nothing was pushed anywhere.
- earlier live runs (pre-restructure): php → clean signed PR here + semgrep-php release PR; python → correctly `failed` (v0.25 `:error` corpus case)
- fix-semgrep-grammar skill against a real breaking bump, locally: `CANNOT_PROCEED` → `failed`, upgrade aborted
- not yet: a real (non-dry) run of the current tip's push + release path — first checklist item

Before enabling:
- [ ] real dispatch of current tip on one language (php)
- [ ] vendor approval for Cursor before flipping `run_review_agent` on
- [ ] branch protection: allow bot-pushed `grammar-update/*`

Note: `GITHUB_TOKEN`-created PRs don't trigger Actions workflows; CircleCI still runs via webhook.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/propose-grammar-update
```diff
@@ -0,0 +1,797 @@
+#!/usr/bin/env python3
```
malaverdiere, line 1: I think you can reuse more of update-grammar since it has a dry run mode. Just my two cents.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+    ).stdout.strip()
+    if not url:
+        return None
+    return url.replace("git@github.com:", "https://github.com/")
+
+
+def latest_stable_tag(url: str) -> str | None:
+    """Return the highest stable release tag at `url`, or None.
+
+    Lists remote tags, keeps only `_STABLE_TAG_RE` matches (so prereleases
```
malaverdiere, line 223: We have to be careful with versions having a dependency on tree-sitter >= 0.25 because of ABI 15.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+         "--ref", tag, "--skip-release"],
+        cwd=root, text=True, capture_output=True, check=False,
+    )
+
+
+def regenerate_snapshots(root: Path, wrapper: str) -> None:
+    """Regenerate corpus snapshots for every grammar in the wrapper.
+
+    A grammar bump almost always churns parse trees, so the committed
+    corpus snapshots go stale. `tree-sitter test --update` rewrites them
```
malaverdiere, line 360: That's cool. I wasn't aware of that. Do you think it'd be useful in the skill I created?

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+    """Branch for a lang+tag proposal.
+
+    Keyed on the exact tag (not the CI user) so re-runs on the same
+    schedule are idempotent and a new upstream tag yields a new branch.
+    """
+    return f"grammar-update/{language}/{tag}"
+
+
+def proposal_exists(root: Path, branch: str) -> bool:
+    """True if a branch (local or on origin) already exists for this proposal.
```
malaverdiere, line 435: That's clever.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+    log_section = (
+        f"<details><summary>test-lang output (tail)</summary>\n\n"
+        f"```\n{log_tail}\n```\n</details>"
+        if log_tail else ""
+    )
+    adapted_note = (
+        "> ⚠️ **An agent adapted existing tests** for this bump (test-lang "
+        "failed on the raw bump, typically a mechanical upstream CST rename). "
+        "Scrutinize the test changes in the diff.\n\n"
+        if result.get("tests_adapted") else ""
```
malaverdiere: I much prefer strongly-typing, like a @dataclass with bool fields in this case.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+        log(f"\n=== would create branch: {branch}")
+        log(f"=== would create PR: {title}\n")
+        print(body)
+        return {**result, "status": "dry-run", "branch": branch}
+
+    # Fork the branch from a CLEAN base (origin/main), NOT the current CI
+    # branch — otherwise the PR carries every commit that branch is ahead of
+    # main. Fetch the base first (CI checks out only the trigger branch).
+    base = base_branch()
+    run_quiet(["git", "fetch", "origin", base], cwd=root)
```
malaverdiere: You use a lot of git operations. Maybe we reach the point that using `uv` with the python git API is worth it.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+        f"In this repo (ocaml-tree-sitter-semgrep), the tree-sitter-{language} "
+        f"submodule was just bumped to {tag} and `lang/test-lang {language}` "
+        f"now FAILS. Review the grammar update and adapt the EXISTING tests so "
+        f"they reflect the new grammar — do not invent new feature tests (a "
+        f"version bump never changes our patching grammar.js).\n\n"
+        f"Typical cause: an upstream rename/restructure of CST nodes that the "
+        f"corpus snapshots in `{corpus}` (and any inherited base-grammar "
+        f"tests) no longer match. Update those expected snapshots to the new, "
+        f"correct shapes. Verify by re-running `cd lang && ./test-lang "
+        f"{language}` until it passes. Only adapt existing test expectations; "
```
malaverdiere: Here we can use the skill. We just need to add a symlink in .cursor

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+
+def agent_cmd(prompt: str) -> list[str]:
+    """Build the Cursor `agent` command for headless CI execution.
+
+    `-p` is print/non-interactive mode; `--force` auto-approves file edits
+    and command execution (without it the agent stops at an interactive
+    permission prompt and does nothing in CI). `--output-format json` makes
+    the result parseable. Model overridable via CURSOR_MODEL.
+    """
+    cmd = ["agent", "-p", "--force", "--output-format", "json"]
```
malaverdiere, line 583: Is there a way to limit execution time / iterations? We could be stuck in a loop for a while.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+
+    Bumps to the latest tag, regenerates snapshots, re-runs test-lang, and
+    classifies the outcome. On a test-lang failure, a Cursor agent reviews
+    the grammar diff and adapts the existing tests, then test-lang is
+    re-run (the authoritative signal). Always resets the working tree
+    before returning (the PR path re-applies changes from the result).
+    Does NOT create a PR — that's the caller's job under --open-pr.
+    """
+    language = target["language"]
+    result: dict[str, Any] = {
```
malaverdiere: Again about the strong typing

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+        result["new_sha"] = new_sha
+
+        if new_sha == old_sha:
+            return {**result, "status": STATUS_NO_OP}
+
+        # Snapshots are expected to churn; regenerate then re-test for the
+        # authoritative result. update-grammar already ran test-lang once
+        # (it may have failed on snapshot drift) — the regen + re-run is
+        # what decides updated-vs-failed.
+        regenerate_snapshots(root, target["wrapper"])
```
malaverdiere: I see this called 3x in the script and that doesn't sound right.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+def downstream_repo_exists(dialect: str) -> bool:
+    """True if a semgrep-<dialect> repo exists to release into.
+
+    Not every test-lang dialect has its own downstream repo: php fans out
+    to [php, php-only] for testing, but only semgrep-php exists (no
+    semgrep-php-only); likewise sfapex has no repo (only semgrep-apex).
+    Releasing a dialect with no repo would fail, so we filter to the ones
+    that exist.
+    """
+    org = os.environ.get("GITHUB_ORG", "semgrep")
```
malaverdiere: I think that update-grammar takes this in consideration.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+        sys.stderr.write(last.stderr)
+        if last.returncode != 0:
+            return last
+    return last  # type: ignore[return-value]
+
+
+DOWNSTREAM_BLOCKED_LABEL = "downstream-blocked"
+
+
+def agent_prompt(language: str, wrapper: str, branch: str, tag: str,
```
malaverdiere: Oh wow. You're pushing the update of semgrep-proprietary from here!
I guess that there's no 'right' or 'wrong' place to do it. But my instinct is to isolate everything in its own repo until Yosef merges everything.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+
+    by_status: dict[str, list[dict[str, Any]]] = {}
+    for r in rows:
+        by_status.setdefault(r.get("status", "unknown"), []).append(r)
+
+    order = [STATUS_UPDATED, STATUS_EXISTS, STATUS_NO_OP, STATUS_NO_TAGS, STATUS_FAILED]
+    icon = {STATUS_UPDATED: "✅", STATUS_EXISTS: "↩️", STATUS_NO_OP: "➖",
+            STATUS_NO_TAGS: "🏷️", STATUS_FAILED: "❌"}
+
+    print("## Grammar update proposals\n")
```
malaverdiere, line 903: I don't know how I feel about running this for many languages. Updating one could get stuck and end up messing up the update for everybody else.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1006 @@
...
+
+def run_update_grammar(language: str, ts_version: str, tag: str,
+                       root: Path) -> subprocess.CompletedProcess:
+    """Bump `language` to `tag` via `update-grammar --skip-release`.
+
+    Subprocessed (not imported) so a failure is isolated from this
+    process and so we capture its combined output for the PR body / log.
+    --skip-release guarantees nothing is pushed to semgrep-<lang>.
+    """
+    return subprocess.run(
```
malaverdiere, line 349: What if the update-grammar script could successfully update everything because the tests pass? Do you have a way to avoid re-testing, etc.?

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1011 @@
...
+        if test.returncode == 0:
+            result["status"] = STATUS_UPDATED
+            return result
+
+        # test-lang failed. Most often this is a mechanical upstream CST
+        # rename that broke existing corpus snapshots — invoke the in-repo
+        # agent to review + adapt those tests, then re-run test-lang as the
+        # authoritative signal.
+        review_log = test.stdout + test.stderr
+        run_review_agent(root, language, target["wrapper"], tag, review_log)
```
malaverdiere: **Bug: `NameError` — `tag` is not defined in this scope.**

`propose()` only has `target["tag"]` / `result["new_tag"]` in scope; there's no local `tag`. This line will raise `NameError` on every bump where the first `test-lang` run fails — which is exactly the "mechanical CST rename" case the review-agent path exists to handle. The crash propagates uncaught out of `main()` (only `KeyboardInterrupt` is caught), so the run fails with a raw traceback instead of a clean `STATUS_FAILED` JSON result.

None of the 22 new unit tests exercise this branch (`test.returncode != 0`), so it shipped untested. Should be `target["tag"]`.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1011 @@
...
+    run_quiet(["git", "fetch", "origin", base], cwd=root)
+    run_quiet(["git", "checkout", "-b", branch, f"origin/{base}"], cwd=root)
+    run_update_grammar(language, target["ts_version"], tag, root)
+    regenerate_snapshots(root, target["wrapper"])
+    # propose() reset the tree, so if it needed the agent to adapt tests,
+    # re-run that here too — otherwise the committed branch would carry the
+    # un-adapted (broken) tests. test-lang already validated this path.
+    if result.get("tests_adapted"):
+        run_review_agent(root, language, target["wrapper"], tag,
+                         result.get("test_log_tail", ""))
```
malaverdiere: **Bug: second review-agent call is primed with the wrong (passing) log and its output is never re-validated.**

By the time `propose()` returns with `tests_adapted=True`, `result["test_log_tail"]` has been overwritten with the *retest* output (line ~613: `result["test_log_tail"] = tail(retest.stdout + retest.stderr)`) — i.e. a **passing** `test-lang` run. This code then re-invokes `run_review_agent` with that passing log, which gets embedded in the prompt as `"Failing test-lang output:\n{tail(test_log, 60)}"` (see `review_agent_prompt`). The agent is told tests are failing when the log shows they pass, so it has no real signal to act on.

Worse, after this second call there's no `test-lang` re-run at all before the branch is committed/pushed/opened as a PR — unlike the first call in `propose()`, which *is* validated by a retest. So the actual committed branch content depends on an unvalidated, independently-run LLM call. The comment above ("test-lang already validated this path") assumes re-running the same prompt reproduces the same validated fix, which doesn't hold given the wrong log and the missing re-test.

(Currently masked by the `NameError` at line 619 — `tests_adapted` can never actually become `True` today — but this becomes live the moment that bug is fixed.)

.github/workflows/propose-grammar-updates.yml
```diff
@@ -0,0 +1,285 @@
...
+        id: gate
+        run: |
+          f="result/result-${{ matrix.language }}.json"
+          if [ -f "$f" ] && [ "$(python -c "import json,sys; print(json.load(open(sys.argv[1])).get('status'))" "$f")" = "updated" ]; then
+            echo "go=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "go=false" >> "$GITHUB_OUTPUT"
+          fi
+      - if: steps.gate.outputs.go == 'true'
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
malaverdiere, line 228: This checkout has no `ref:`, so it checks out the default trigger ref (`main` for both `schedule` and the typical `workflow_dispatch`) — not the `grammar-update/<lang>/<tag>` branch the `propose` job pushed.

`run_release_and_trigger`'s own docstring says: "Assumes the `grammar-update/<lang>/<tag>` branch already exists in this repo ... and is checked out." For the common case this happens to be harmless (the re-bump from `main` reproduces the same generated content deterministically), but for any language where a Cursor review-agent adapted the tests (`tests_adapted=True`), those adapted corpus/test fixes only exist on the pushed branch, not on `main` — a plain re-bump from `main` won't reproduce them, so `lang/release` could publish content that was never actually validated by `test-lang`. Should this job check out `grammar-update/${{ matrix.language }}/<tag>` explicitly (or download the branch state as an artifact) instead of relying on the default ref?

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1011 @@
...
+def run_release_and_trigger(root: Path, target: dict[str, Any],
+                            dry_run: bool) -> int:
+    """Downstream hand-off: release to semgrep-<lang>, then launch the agent.
+
+    Assumes the grammar-update/<lang>/<tag> branch already exists in this
+    repo (pushed by the propose job) and is checked out. Re-applies the
+    bump locally so `lang/release` has the generated sources, releases to
+    the downstream branch, and triggers the proprietary-PR agent.
+    """
+    language, wrapper, tag = target["language"], target["wrapper"], target["tag"]
```
malaverdiere: `target["tag"]` here comes from a fresh `resolve_target()` call in this (separate) `--release-and-trigger` invocation — i.e. a brand-new `git ls-remote` against upstream — rather than the tag `propose()` actually bumped-to/tested/PR'd (which isn't threaded through the `result-*.json` artifact). If upstream publishes a newer stable tag in the window between the `propose` and `integrate` jobs of the same workflow run, `branch_name(language, tag)` computes a branch name that doesn't match the one `open_pr()` created, and/or this releases a version that was never actually run through `test-lang`. Narrow window, but worth passing the validated tag through the artifact instead of re-resolving it live.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1011 @@
...
+        print(body)
+        return {**result, "status": "dry-run", "branch": branch}
+
+    # Fork the branch from a CLEAN base (origin/main), NOT the current CI
+    # branch — otherwise the PR carries every commit that branch is ahead of
+    # main. Fetch the base first (CI checks out only the trigger branch).
+    base = base_branch()
+    run_quiet(["git", "fetch", "origin", base], cwd=root)
+    run_quiet(["git", "checkout", "-b", branch, f"origin/{base}"], cwd=root)
+    run_update_grammar(language, target["ts_version"], tag, root)
```
malaverdiere: Efficiency note: `propose()` already runs a full bump + snapshot regen + `test-lang` (potentially twice, once more if the review agent fires) and then discards it via `reset_language_state`. `open_pr()` then redoes `run_update_grammar` + `regenerate_snapshots` from scratch to produce the real commit — so every `updated` language pays for the expensive OCaml build/parse cycle twice per job, and for `tests_adapted` languages a *third* Cursor agent invocation happens here with no re-validation (see the comment on line 468). If this duplication is an intentional tradeoff for a clean base branch (as the comment above suggests), it'd be worth calling out explicitly in the PR description so reviewers know the cost is expected.

malaverdiere (commented):
Reviewed the propose/integrate pipeline. Left inline comments on the specific issues found (two confirmed correctness bugs in `scripts/propose-grammar-update`, a checkout/branch mismatch in the workflow, a tag-resolution race, and an efficiency note on the duplicated bump/regen cycle).

One additional point that doesn't map to a single line: `run_review_agent` / `trigger_agent` shell out to the Cursor `agent` CLI, sending grammar source, failing `test-lang` logs, and (via the downstream hand-off) `semgrep-proprietary` content to Cursor's cloud service. Per org policy the currently-approved third-party services are AWS, DataDog, Sentry, Salesforce, and OpenAI — Cursor isn't on that list. The PR's own "before enabling" checklist already flags confirming Cursor's access to `semgrep-proprietary` as outstanding, but the code path fires unconditionally once `CURSOR_API_KEY` is set, with no vendor-approval gate. Worth explicitly resolving (go/vendor-onboarding or #ask-techops) before this is turned on for real.

scripts/propose-grammar-update
brandonspark: Fixed in bbd6220 — bound at the top of propose(), and the failure branch has tests now.

scripts/propose-grammar-update
brandonspark: Fixed in bbd6220 by restructuring: propose() does bump/regen/test/skill once on the final branch and commits it; open_pr just pushes. The second agent call is gone entirely.

.github/workflows/propose-grammar-updates.yml
brandonspark: The release step now fetches and checks out grammar-update/<lang>/<tag> itself (bbd6220), and as of 5f48f07 the tag/branch come from the propose job's artifact — so the default ref here no longer matters.

scripts/propose-grammar-update
brandonspark: Closed in 5f48f07 — integrate reads the validated tag/branch from result-<lang>.json (--result) instead of re-resolving.

scripts/propose-grammar-update
brandonspark: Gone with the bbd6220 restructure — the expensive cycle runs once, on the PR branch.

scripts/propose-grammar-update
brandonspark: Done in 5f48f07 — Target/Result dataclasses; the result artifact round-trips through Result.to_json()/from_file().

scripts/propose-grammar-update
brandonspark: Same — Result dataclass as of 5f48f07.

scripts/propose-grammar-update
brandonspark: Done in 689800a — it drives the skill now, and the .cursor/skills symlink is in.

scripts/propose-grammar-update
brandonspark: Added in 98bb63f: timeout-minutes on every job plus a 25m timeout wrap on the agent call.

scripts/propose-grammar-update
brandonspark: Right — fixed in bbd6220. Called once now: keep-mode propose commits the validated tree and open_pr just pushes.

scripts/propose-grammar-update
brandonspark: Agreed, dropped in 5f48f07. This now stops at the semgrep-<lang> PR; proprietary integration stays with the existing Cursor workflow over there.

scripts/propose-grammar-update
brandonspark: This flow never touches the version pins — only the submodule moves. A grammar that starts requiring >= 0.25 fails test-lang loudly rather than changing ABI silently.

scripts/propose-grammar-update
brandonspark: fail-fast is off, every job has a timeout, and each language is its own branch/PR, so a stuck one can't block or corrupt the rest. The pre-filter also keeps up-to-date languages out of the matrix entirely.

scripts/propose-grammar-update
brandonspark: Different layer I think: LANGUAGE_VARIANTS handles the version-list files, this filters to dialects that actually have a semgrep-<x> repo (php-only and sfapex don't), which only matters at release time.

scripts/propose-grammar-update
brandonspark: Maybe as a first draft — the catch is --update happily bakes in a wrong tree, so the skill would have to review the rewritten expectations rather than trust them. As-is it'd fight the skill's 'verify against the actual tree' rule.

scripts/propose-grammar-update
brandonspark: Not currently — update-grammar runs test-lang itself even under --skip-release, and we re-run after regenerating snapshots, so one run is wasted per bump. A --skip-test flag on update-grammar would fix it; happy to do that as a follow-up.

scripts/propose-grammar-update
brandonspark: It imports update-grammar's helpers and subprocesses it for the bump now; what's left here is mostly the orchestration it adds (branching, PR, artifact). The one real overlap left is the double test-lang — see the thread below.

scripts/propose-grammar-update
brandonspark: Probably follow-up territory — stdlib-only keeps the CI job dependency-free. Agreed it's worth it if the git surface keeps growing.

brandonspark:
Re: Cursor scope — as of 5f48f07 the proprietary stage is gone entirely, and the review agent is opt-in via the run_review_agent dispatch input (off by default, including any scheduled run). Without it the CLI is never installed and CURSOR_API_KEY is never exported to the job, so nothing can leave the runner. Turning the flag on for real still needs vendor onboarding — called out in the description's enable checklist.

scripts/propose-grammar-update
malaverdiere: Okay. Long term, we may just end up merging them anyways.

scripts/propose-grammar-update
malaverdiere: Then maybe we want to avoid using the --update command?

malaverdiere:
But it would work locally, right?

brandonspark:
> But it would work locally, right?

what's this mean? we were able to test it locally with a broken grammar update + using your fix grammar script, to see if we should proceed to push the `semgrep-<lang>` change

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+    """
+    valid = ug.discover_valid_languages(root)
+    wrapper = ug.validate_language(language, valid)
+    submodule = ug.submodule_path(root, language)
+    url = gitmodules_url(root, submodule)
+    # Languages absent from every languages-* file (e.g. gomod, a "pro"
+    # package handled outside the standard version scheme) aren't
+    # proposable via this flow; surface that as skip_reason rather than
+    # dying, so callers can record it and move on.
+    ts_version = language_ts_version(root, wrapper)
```
malaverdiere, line 267: We want to ideally update the tree-sitter version to something newer while we're at it. Some grammar updates may require building with a more up-to-date tree-sitter version.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+    # duplication (and the second, unvalidated agent run it caused).
+    branch = branch_name(language, tag)
+    committed = False
+    if keep:
+        base = base_branch()
+        run_quiet(["git", "fetch", "origin", base], cwd=root)
+        run_quiet(["git", "checkout", "-B", branch, f"origin/{base}"], cwd=root)
+
+    try:
+        bump = run_update_grammar(language, target.ts_version, tag, root)
```
malaverdiere, line 665: Success is not checked.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+    body = pr_body(result, target.url)
+
+    if dry_run:
+        log(f"\n=== would push branch: {branch}")
+        log(f"=== would create PR (base {base}): {title}\n")
+        log(body)  # stderr: stdout must stay JSON-pure for the artifact
+        result.status = "dry-run"
+        return result
+
+    if proposal_exists(root, branch):
```
malaverdiere, line 536: Shouldn't that be checked earlier on?

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+
+    Reads the committed gitlink via `git ls-tree` — no submodule checkout
+    needed, so this is cheap enough for the enumerate pre-filter.
+    """
+    rel = submodule.relative_to(root).as_posix()
+    out = git_query(["git", "ls-tree", "HEAD", rel], cwd=root).stdout.split()
+    return out[2] if len(out) >= 3 else None
+
+
+def tag_commit_sha(url: str, tag: str) -> str | None:
```
malaverdiere, line 294: returns the wrong SHA for annotated tags. The tag-object SHA sorts before the peeled `^{}` commit SHA, so `.split()[0]` grabs the wrong one.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+
+def run_quiet(cmd: list[str], cwd: Path) -> None:
+    """Run a mutating command, mirroring its output to STDERR, raising on error.
+
+    stdout is reserved for the single JSON result line, so git/gh output
+    must never go there (it would corrupt the contract and the workflow's
+    `tee`'d artifact). Captures and re-emits on stderr.
+    """
+    log(f"$ {' '.join(cmd)}  (in {cwd})")
+    r = subprocess.run(cmd, cwd=cwd, text=True, capture_output=True, check=False)
```
malaverdiere: I think that the subprocess library has a way to combine the output to stderr for you. Unless you wanted to separate them? If so, I suggest that you also add a banner that identifies which is which.

scripts/propose-grammar-update
```diff
@@ -0,0 +1,1025 @@
...
+
+
+def latest_stable_tag(url: str) -> str | None:
+    """Return the highest stable release tag at `url`, or None.
+
+    Lists remote tags, keeps only `_STABLE_TAG_RE` matches (so prereleases
+    and non-version tag schemes are ignored), and returns the highest by
+    numeric version.
+    """
+    result = git_query(["git", "ls-remote", "--tags", "--refs", url], cwd=Path.cwd())
```
malaverdiere, line 227: My only concern is that we could get a version that depends on tree-sitter ABI 15, which we're not really ready for.

malaverdiere (approved):
`tag_commit_sha` should be addressed before merging. The rest is something to implement soonish

scripts/propose-grammar-update
brandonspark: Good catch — fixed in 9d8e9ea: prefer the peeled ^{} line, fall back to the plain ref for lightweight tags. Unit-tested both. (Most grammar repos use lightweight tags, so the pre-filter list barely changes in practice, but annotated ones would've looked permanently behind.)

scripts/propose-grammar-update
brandonspark: Fixed in 9d8e9ea. Nuance: a non-zero exit with a MOVED submodule is expected — update-grammar's own final test-lang fails on the stale snapshots we regenerate right after — so only the unchanged-submodule case treats it as failed (it was silently becoming no-op before).

scripts/propose-grammar-update
brandonspark: Yes — 9d8e9ea adds the check before the build (existing_proposal in main). open_pr still re-checks right before pushing to cover the race.

scripts/propose-grammar-update
brandonspark: Done in 9d8e9ea — stderr=STDOUT at capture time, so the re-emitted log keeps the real interleaving.

scripts/propose-grammar-update
brandonspark: I'd keep it here — it plays a different role than in the skill. Here --update generates the diff a human reviews: the full corpus diff lands in the PR body, :error cases are never auto-updated, and parse-examples still gates real files. Without it, any bump with tree churn fails and needs hand-editing, which is the thing this automates. In the skill, --update would replace verification; here it feeds it.

scripts/propose-grammar-update
brandonspark: Agreed, but I'd do it as a follow-up — update-grammar already takes the version as an argument, so the workflow could grow a ts_version input that also moves the languages-* lists. This PR deliberately keeps the pins fixed so a bump can only change the submodule.

scripts/propose-grammar-update
brandonspark: It can't arrive via this pipeline: parser.c is regenerated by OUR pinned CLI (that's the PATH setup in the workflow), so the output ABI is fixed by the pin — upstream's checked-in parser.c is never consumed. A grammar.js that needs newer generator features fails tree-sitter generate, test-lang goes red, status=failed, no PR. ABI 15 only enters when someone moves the pins, which this flow never does.

PR #615, opened 2026-06-25
Bump core submodule to latest main (ABI mismatch fix)

malaverdiere:
## Summary

Updates the `core` submodule (`ocaml-tree-sitter-core`) in include:

* #128 "fix(tree-sitter): fail loudly on parser/runtime ABI mismatch": the generated per-language parser stubs now check the return value of `ts_parser_set_language` and raise an OCaml exception on an ABI-version mismatch, instead of discarding it and silently producing empty/`null` parse trees.

Note that this does NOT regenerate any grammars' wrapper to use the new template. As we're expecting many grammar updates to happen soon, I prefer to keep this PR simple.

## Testing

- `make test`
- `make -C core e2e`
- `make test-python`


nmote (approved):

malaverdiere:
### Merge activity

* **Jun 30, 4:32 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/615).
* **Jun 30, 4:32 PM UTC**: @malaverdiere merged this pull request with [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/615).

PR #616, opened 2026-06-29
bug: Use stamped tree-sitter version for fyi/tree-sitter-version

malaverdiere:
## What

`lang/release` now reads the tree-sitter version out of the **provenance comment on the first line of the generated** **`parser.c`** to populate `fyi/tree-sitter-version`, instead of invoking the global `core/tree-sitter` symlink.

## Why

The old code used the `core/tree-sitter/bin/tree-sitter --version` output to generate `fyi/tree-sitter-version`, regardless of what was used to generate the grammar. Sourcing it from the shipped parser keeps the stamp accurate and self-consistent with the code being released.

## Dependency (must merge together)

This relies on the provenance comment written by:

- **semgrep/ocaml-tree-sitter-core#129** — https://github.com/semgrep/ocaml-tree-sitter-core/pull/129

### Checklist

- [ ] Any new parsing code was already published, integrated, and merged into Semgrep. DO NOT MERGE THIS PR BEFORE THE SEMGREP INTEGRATION WORK WAS COMPLETED.
- [x] Change has no security implications (otherwise, ping the security team)

This relates to LANG-575 but does NOT close it.

PR #617, opened 2026-07-01
bug: release to abort on grammar submodule drift

malaverdiere:
## What

`lang/release` now **fails fast** if a grammar-source submodule (e.g. `tree-sitter-dart`) is checked out at a commit **different from the one pinned in git**, instead of silently generating from — and stamping `fyi/versions` with — that uncommitted commit.

We deliberately **do not auto-sync** — drift means someone either forgot to commit a new pin or has a stale checkout, and only the human can decide which.

## Why

`fyi/versions` records each grammar's commit from the submodule's *checked-out* `HEAD`. If the working tree has drifted from the pinned pointer, the recorded upstream commit (and the grammar actually generated) can silently disagree with what's committed.

## Scope / behavior notes

- Checks only the submodules referenced by the released language's `fyi.list`.
- Runs in `--dry-run` too, so previews are trustworthy.

### Checklist

- [x] N/A Any new parsing code was already published, integrated, and merged into Semgrep. DO NOT MERGE THIS PR BEFORE THE SEMGREP INTEGRATION WORK WAS COMPLETED.


malaverdiere:
* **#620** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/620?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#617** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#616** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

nmote:
The Python script seems like overkill and doesn't really pass the smell test for me. This feels like something that should be a ten line bash script, if that. We should probably generally check for a dirty working directory on release and fail on uncommitted changes (if we don't already). Any dirty submodules will show up in `git status`. The only exception I'm aware of is if they have never been initialized. We should check for that too, but it's hard to imagine that being more than a couple lines in Bash.

malaverdiere:
It's funny because I had Brandon complaining about new bash scripts when they could be Python. But you're right that there's a point where Python adds a lot of verbosity for the same task

nmote:
My objection isn't to Python specifically, just that this particular Python script seems overengineered. Bash is more concise and for something as simple as I believe this is, I think it's fine, but I'd also be fine with that simple logic getting implemented in Python even though it would probably be maybe twice the lines of code. Still, I'd expect something on the order of 10-20 lines of Python code to accomplish this, unless I have fundamentally misunderstood the problem.

PR #622, opened 2026-07-09
Use uv and GitPython in grammar-update scripts

malaverdiere:
## Summary

- Run Python scripts/tests via `uv` (`pyproject.toml` + `uv.lock`) instead of ad hoc pip/`python`
- Drive git ops in `update-grammar` / `propose-grammar-update` with GitPython
- Run pre-commit before staging so auto-fixes land before commit; harden `ensure_on_branch` for detached HEAD

## Test plan

- [ ] `make test-python`
- [ ] Spot-check propose workflow jobs still invoke scripts via `uv run`


malaverdiere:
Downstream grammar PR:
- https://github.com/semgrep/semgrep-ruby/pull/3

PR #626, opened 2026-07-14
fix: remove javascript grammar

tean-lai:
Resolves LANG-592.

Semgrep uses the typescript tree-sitter parser for javascript, so the semgrep-javascript grammar is outdated. This PR removes it.

In the future:
- semgrep-javascript is deprecated
- javascript is considered a sublang of typescript

### Checklist


lang/semgrep-grammars/lang/javascript
```diff
@@ -1 +1 @@
-../src/semgrep-javascript
\ No newline at end of file
+../src/semgrep-typescript/javascript
```
nmote, line 1: Probably less confusing to just remove this symlink altogether

nmote (approved):
Probably also worth archiving https://github.com/semgrep/semgrep-javascript

lang/semgrep-grammars/lang/javascript
tean-lai: The plan was to archive semgrep-javascript, but I seem to have insufficient perms at the moment.

Keeping the symlink seems to be more consistent with other stuff. Thinking particularly of `tsx`, which we treat as a sublang of typescript already, but that similarly symlinks to `semgrep-typescript/tsx`. It also seems like the `test-lang` script depends on the existence of this symlink.

malaverdiere:
## Review: JavaScript-grammar removal

Checked whether automated tests still pass, whether the TypeScript grammar is affected, and swept for leftover JavaScript references. Verified all of this by actually running the scripts locally (initialized the `tree-sitter-javascript`/`tree-sitter-typescript` submodules, ran `prep`, ran the vendored `tree-sitter` 0.22.6 binary directly), not just by reading the diff.

### ✅ Already fixed by 8693d43 ("fix corpus symlink for javascript")
`lang/semgrep-grammars/src/semgrep-typescript/prep` originally pointed `javascript/corpus/inherited` at `../../../tree-sitter-javascript/javascript/corpus`, which doesn't exist (tree-sitter-javascript is a flat repo — corpus lives at `test/corpus`). That's been corrected to `../../../tree-sitter-javascript/test/corpus`. Confirmed the symlink resolves now. (For what it's worth, I also confirmed this specific corpus-test path is inert either way today — `tree-sitter test` only reads `test/corpus`, not bare `corpus`, and these sublang dirs use `corpus/` directly — so this was latent either way, not an active regression. Still, good to have it correct.)

### 🔴 Still open — will break CI
**`.circleci/config.yml:159`** still lists `- javascript` in the `test-language` matrix. That job runs `./test-lang javascript`, whose very first line is `cd semgrep-grammars/src/semgrep-javascript` — a directory this PR deletes (`c9e95e5 remove semgrep-javascript`). I reproduced the failure locally:
```
$ cd lang && cd semgrep-grammars/src/semgrep-javascript
cd: semgrep-grammars/src/semgrep-javascript: No such file or directory
```
There's no `allow_failure`/`continue-on-error` on this job, so it'll hard-fail on every push. `lang/Makefile`'s own coupling comment says any `SUPPORTED_TS_LANGUAGES` change must be mirrored here — `javascript` was correctly dropped from the Makefile lists but the CircleCI matrix entry wasn't. Since javascript is now a sublang of typescript (like `tsx`, which also isn't listed separately here), it should just be removed from this matrix.

### 🟡 Minor — stale JavaScript references left behind
- **`doc/fix-semgrep-grammar-skill.md:9` and `:24`** still use `javascript` as the `test-lang` example (`e.g. javascript`, `test-lang javascript pass`), while the canonical `.agents/skills/fix-semgrep-grammar/SKILL.md` had its matching examples swapped to `typescript` in this PR (09c990f). Worth syncing so the two docs don't drift.
- **`.agents/skills/fix-semgrep-grammar/SKILL.md:215`** still says "semgrep-javascript's `_jsx_child`" as a worked example — that directory is gone; the rule now lives at `lang/semgrep-grammars/src/semgrep-typescript/javascript/grammar.js`. Worth rewording so it doesn't point at a directory that no longer exists.

### TypeScript grammar
Unaffected. The `prep` script's typescript/tsx blocks are untouched, and none of the javascript-related additions can abort the script early (no early exit before the typescript/tsx blocks run).

PR #631, opened 2026-07-17
impl slack notifications with fable 5

tean-lai:
When the grammar automation script fails, we well alert #team-languages on Slack. You could also trigger a failure, as well as redirect the notification to DM a specific user on Slack.

This is implemented at a GitHub workflows level as opposed to on a script level. This does have the advantage of working on failures beyond the scope of a script. 

### Testing
For injecting a failure while testing, consider the following commands:
```
gh workflow run "Propose grammar updates" --ref tean/lang-604-alert-on-slack-when-unable-to-update-a-grammar-v3 -f inject_failure=enumerate -f notify_channel=U08THQM5A01
```
There are two inputs here:
- `inject_failure` (`none` / `enumerate` / `propose` / `integrate`)
- `notify_channel` - a specific channel or Slack member ID

Have done some runs. Verified it works for injecting failures. Also successfully sends notifications on failures, though it sends one notification after we attempt to propose all languages but have at least one failure.

Closes LANG-604.

### Secrets
This relies on the `SLACK_BOT_TOKEN` GitHub secret. It's used to interact with the Slack app named `Grammar Automation Bot`. Ownership of app is under me (Tean) right now.

### Checklist

### DIFF
--- a/.github/workflows/propose-grammar-updates.yml
+++ b/.github/workflows/propose-grammar-updates.yml
@@ -8,6 +8,10 @@ name: Propose grammar updates
 # separate integrate job then releases each green bump to semgrep-<lang> via
 # PR. semgrep-proprietary integration is deliberately out of scope.
 #
+# Any failed job triggers a Slack alert to #team-languages (see the `notify`
+# job; needs the SLACK_BOT_TOKEN secret). Test the alert path end-to-end with
+# the inject_failure + notify_channel inputs.
+#
 # See scripts/propose-grammar-update for the per-language logic.
 
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/propose-grammar-updates.yml
```diff
@@ -306,3 +333,42 @@ jobs:
...
+  # from inside (build breakage, timeouts, crashes). The message carries only
+  # job statuses + a run link — never grammar source or logs.
+  notify:
+    name: Alert on failure
+    needs: [enumerate, propose, integrate]
+    if: ${{ always() && contains(needs.*.result, 'failure') }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
```
malaverdiere, line 353: I think that using the official Slack API would make sense: https://docs.slack.dev/tools/slack-github-action/ remember to pin to a specific git commit instead of a version tag.

I was also thinking about it, and we should also notify in success, with a link to the PR. I don't know if there's a clean way to do this without using structured output like JSON.

In case of failure, does this notification also give a link to the log on github? Because that would be the first thing that the user would want to do.

Lastly, Fable 5 is very expensive. I encourage you to use cheaper models like Opus 4.8 or Sonnet 5. I use Sonnet 5 and I'm fairly satisfied.

tean-lai: On the slack api, did you mean like 419? I thought this was pinning to a commit.

Added the notify on success and linking on failure.

Fable 5 does seem expensive, but I think since this is a smaller script it isn't so bad. The other models seem to suggest changes of hundreds more lines of code, and this seemed like the smallest maintenance burden.

malaverdiere: You pinned a commit, that's fine.

malaverdiere (approved):

PR #633, opened 2026-07-23
fix: remove second local git check on grammar automation

tean-lai:
Basically, when running `propose-grammar-update`, we run into the following problem:
1. Script tries to propose a PR with name `grammar-update/LANG/version`
2. Script checks there is no local git branch with the same name. Aborting if so.
3. If it can proceed, it'll create a local git branch with the name.
4. Then, it tries to open the pr in `open_pr`, but repeats step #2, always leading to abortion since we created said branch in #3.

Tldr; we never open a pr for grammar automation

This fix addresses that by removing the second check for a conflicting local branch.

### Testing
Still needs to be tested, one way to go about this is to run the workflow on both this branch and the main branch, and compare.

### Checklist

### DIFF
--- a/scripts/propose-grammar-update
+++ b/scripts/propose-grammar-update
@@ -945,11 +945,6 @@ def open_pr(root: Path, target: Target, result: Result,
         result.status = "dry-run"
         return result
 
-    if proposal_exists(root, branch):
-        result.status = STATUS_EXISTS
-        result.detail = f"branch {branch} already exists"
-        return result
-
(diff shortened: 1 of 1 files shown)


scripts/propose-grammar-update
malaverdiere: No longer true

malaverdiere: no longer true

malaverdiere (approved):
Tweak docs otherwise OK

PR #637, opened 2026-07-24
fix: make propose-grammar-update crash-resilient, always emit Result JSON

malaverdiere:
## Summary

- `ug.die()` now raises `SystemExit(f"error: {msg}")` instead of printing and calling `sys.exit(1)` directly, so it can be caught like any other exception without losing the message.
- `propose-grammar-update`'s `main()` wraps the language-specific work in a `try`/`except SystemExit`/`except Exception`: any crash after the language is known (a `die()`, a `git`/`gh` failure, anything unexpected) still prints a `status: failed` Result JSON naming the language and logs the detail to stderr, instead of leaving no artifact and a bare traceback. `propose()` itself has the same guarantee for crashes during the bump/test-lang/commit sequence.
- The "Propose update" workflow step keeps `| tee "result-<lang>.json"` (with `set -o pipefail`, since bash's default `-e` doesn't include it) so the result JSON — including the failure log tail — stays visible live in the Actions log, not just in the uploaded artifact.
- `summarize()` / `load_result_row()` turn empty or corrupt result artifacts (e.g. from a hard kill) into failed rows instead of silently dropping them, so the summary table and Slack alert always name every language that ran.
- The "Skip unless updated" gate and the Slack "Compose message" step were hardened so one missing/corrupt `result-<lang>.json` can't abort the whole step (per-file jq loop, defensive JSON parsing).

This closes LANG-621.

## Test plan

- [x] `python -m unittest scripts.test_propose_grammar_update`
- [x] `python -m unittest scripts.test_update_grammar` (for `ug.die` / `SystemExit.code`)
- [x] Workflow dry-run / `inject_failure` still surfaces failed propose jobs and the Slack alert


scripts/propose-grammar-update
```diff
@@ -1341,21 +1347,65 @@ def list_languages(root: Path, as_json: bool, updatable_only: bool = False) -> N
         print("\n".join(names))
 
 
+_RESULT_FILENAME_RE = re.compile(r"^result-(.+)\.json$")
+
+
+def language_from_result_path(path: Path) -> str | None:
```
malaverdiere, line 1351: The script should have the language by the point the failure occur. Don't try to guess from the paths.

scripts/propose-grammar-update
```diff
@@ -1341,21 +1347,65 @@ def list_languages(root: Path, as_json: bool, updatable_only: bool = False) -> N
...
+        return {"language": fallback_lang, "status": STATUS_FAILED,
+                "detail": "invalid result JSON: expected an object"}
+    if not r.get("language"):
+        r["language"] = fallback_lang
+    if not r.get("status"):
+        r["status"] = "unknown"
+    return r
+
+
+def _is_valid_result_file(path: Path) -> bool:
```
malaverdiere: Is there a way to avoid getting a malformed json file in the first place? In that way that would be minimally invasive?

brandonspark (approved):

malaverdiere:
* **#638** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/638?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#637** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/637?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/637?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

malaverdiere:
### Merge activity

* **Jul 25, 12:36 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/637).
* **Jul 25, 12:36 AM UTC**: @malaverdiere merged this pull request with [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/637).

PR #641, opened 2026-07-27
fix: share grammar-update tree-sitter CI cache across languages

malaverdiere:
Fixes LANG-625: the propose matrix no longer races on one shared tree-sitter cache key. `core/_build` stays keyed by core SHA; tree-sitter pins use one shared key derived from `lang/scripts/ts-versions`, and each job runs `provision-tree-sitter-all` so the first save includes every pin.

Note that the difference in `provision-tree-sitter-all` is adding the executable bit.

### Checklist


.github/workflows/propose-grammar-updates.yml
yosefAlsuhaibani, line 138: I think we can avoid having to roll our own caching mechanism if we added `dune-cache: true` (not sure if that's the exact name of the flag) here to enable the build cache (?)

yosefAlsuhaibani (approved):
Makes sense! added a comment.

.github/workflows/propose-grammar-updates.yml
yosefAlsuhaibani: (only for the `core/_build` cache and not the ts cache ofc)

malaverdiere: `setup-ocaml`’s dune cache key hashes **job name**, so `Propose php` and `Propose ruby` get different primary keys and share only via a broad platform/arch restore fallback, so we would have less efficient caching.

malaverdiere:
### Merge activity

* **Jul 29, 2:59 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/641).
* **Jul 29, 2:59 PM UTC**: @malaverdiere merged this pull request with [Graphite](https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/641).

PR #646, opened 2026-07-29
ci: replace CircleCI language matrix with GitHub Actions (LANG-579)

malaverdiere:
### Checklist

- [x] Any new parsing code was already published, integrated, and merged into Semgrep. DO NOT MERGE THIS PR BEFORE THE SEMGREP INTEGRATION WORK WAS COMPLETED.
- [x] Change has no security implications (otherwise, ping the security team)

## Summary

* Ported the CircleCI `test-language` to GHA
  - Refactored common job configurations
  - Added the `list-languages` utility, which has slightly different requirements than `propose-grammar-updates --list-languages`.
  - Configured a bigger runner for C# pro because it was getting killed on the default Ubuntu one.
* Switched from GCC to Clang because old C++ grammars have a compilation error in GCC.
* Housekeeping:
  - Added CFML to the list of languages because it was missing.
  - Removed hardcoded lists in the makefiles

This closes LANG-579.

## Test plan

- [x] PR CI: **Test languages** aggregate is green
- [x] Existing **ci** workflow still builds/tests core via `build-install-core`

PR #648, opened 2026-07-30
ci: cut the per-job setup cost of the language matrix

malaverdiere:
Cuts the per-job setup cost of the language matrix. Measured, not estimated.

## The problem

On `test-lang lua` (run 30511904902, warm caches) `build-core` took 187s while the actual `test-lang` work took **5s**. The job is ~97% setup, and the matrix pays it 47 times:

| phase | time | share |
|---|---|---|
| `make setup` (opam deps) | 120s | 64% |
| `git submodule update` | 53s | 28% |
| `setup-node` | 7s | 4% |
| caches + provision | 3s | 2% |

## Changes

- **Cache the opam switch.** `setup-ocaml` v3 uses a *workspace-local* switch, so the 42 dependency packages install into `./_opam`, not the opam root. Both paths are cached; `_opam/.opam-switch/build` is excluded since opam does not need it post-install and it dominates the size. Keyed on `core/*.opam` (the dependency set) rather than the core tree SHA, so editing `core/` does not invalidate it.
- **Drop `--recursive` from the submodule init.** The only nested submodules are `tree-sitter-bash`'s copy of `bash-it` plus its four bats test libraries, and two spare checkouts of tree-sitter itself under `tree-sitter-dart/` and `tree-sitter-hcl/fuzz/`. Nothing under `lang/`, `core/scripts` or any `prep` script reads them; the 47 top-level grammar submodules still init.

## Measured outcome

`build-core` duration across all successful matrix jobs:

| variant | median | max | total runner time |
|---|---|---|---|
| baseline | 189s | 221s | ~150 min |
| caching `~/.opam` only | 191s | 275s | 153 min |
| **+ `./_opam`** | **85s** | 216s | **80 min** |

Median setup **-55%**, roughly **70 min of runner time saved per run**. The remaining `max` outliers are the first-wave jobs, which miss the cache and pay the save. Wall clock is unchanged (~22 min) because the matrix is bound by `max-parallel: 8`, not by per-job time.

All 47 languages green: run 30557467211.

## Note on `semgrep/setup-ocaml`

Review feedback on #646 correctly identified that our fork caches deps + the switch, and that diagnosis is what found the bug above. Not switching to it here: the `save-opam-post-run` input that implements this is only on the fork's `master`, not in any release (checked v3.2.18..v3.4.3), and that master is 23 commits behind the `ocaml/setup-ocaml` v3.6.1 this workflow pins. **Once a fork release carries that input, this entire cache block collapses to `save-opam-post-run: true`** and should be replaced.

### DIFF
--- a/.github/actions/build-core/action.yml
+++ b/.github/actions/build-core/action.yml
@@ -10,8 +10,12 @@ runs:
     - name: Rewrite SSH submodule URLs to HTTPS
       shell: bash
       run: git config --global url."https://github.com/".insteadOf "git@github.com:"
-    # --jobs: ~50 grammar submodules, cloned sequentially by default.
-    - run: git submodule update --init --recursive --depth 1 --jobs 8
+    # --jobs: ~50 grammar submodules, cloned sequentially by default. Not
+    # --recursive: the only nested submodules are tree-sitter-bash's copy of
+    # bash-it (plus its four bats test libraries) and two spare checkouts of
+    # tree-sitter itself under tree-sitter-dart/ and tree-sitter-hcl/fuzz/.
+    # Nothing under lang/, core/scripts or any prep script reads them.
+    - run: git submodule update --init --depth 1 --jobs 8
       shell: bash
     - run: ./core/scripts/setup-node
       shell: bash


malaverdiere:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/648?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#648** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#646** <a href="https://app.graphite.com/github/pr/semgrep/ocaml-tree-sitter-semgrep/646?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

malaverdiere:
Absorbed into #646. Once the opam switch caching moved downstack into #646 (via semgrep/setup-ocaml's save-opam-post-run), this PR was down to a single line — dropping `--recursive` from the submodule init — so it has been folded into #646 rather than kept as its own PR. #649 has been rebased onto #646.

PR #660, opened 2026-08-21
fix(dockerfile): parse env ports with protocol in EXPOSE

charle-z:
Fixes the grammar side of semgrep/semgrep#11798.

Docker supports environment replacement in `EXPOSE`, including a valid `/tcp` or `/udp` suffix, but the extended Dockerfile grammar only allowed a numeric `expose_port` with protocol or a standalone expansion. This adds the missing `expansion + protocol` form without broadening accepted protocols.

Coverage includes `$PORT/udp`, `${PORT}/tcp`, mixed multiple `EXPOSE` values, a following `ENTRYPOINT`, and negative cases for unsupported/empty protocol suffixes.

Validation:
- Dockerfile tree-sitter corpus: 138/138 pass
- `make install`: pass
- `cd lang && ./test-lang dockerfile`: pass
- generated C/OCaml parser builds; parser examples 2/2 pass; no undesirable Blank nodes


malaverdiere:
Sorry to keep you waiting. We did a substantial repo consolidation. The [documentation](https://docs.semgrep.dev/contributing/updating-a-grammar) is also live on our website. You can create a PR and tag me directly.
It should be a simple matter of taking the PRs in both repos, rebasing and combining them into a single PR in `semgrep` and regenerating the OCaml wrapper.
If possible, update the tree-sitter version to 0.26.3 for dockerfile. Tree-sitter sometimes makes updates complicated, so don't spend too much time on that if it's not a one-shot.

